PR #6, opened 2022-08-15
ESDS-58 Remove es-bs-extends

tomleo:
- Remove `npm --prefix es-vue-base run build` from `make update`. The command was failing until I installed npm packages.
- Remove `es-bs-extends` (64a28fda0031dbed0b14057d1a6a4661df2fa1e5) and move additions into `es-bs-base` (894ec8638341452e51515ef3256e49a67514f31d)
- Fix snapshot test by passing in date as a argument (dd7e6ed6fa700fc5531dff580818bb586231669c)
- Turn off formatOnSave setting (0e6f7028478666c51bbbdb0ff5290819d513fb30) as it was triggering some format changes that violated package linting rules


Makefile
```diff
@@ -20,5 +22,4 @@ publish:
 
 .PHONY: update
 update:
-	npm --prefix es-vue-base run build
```
RyanMulready, line 23: Why the removal of this? It needs to happen at least the first time before you run `make dev`

es-bs-base/scss/_variables.scss
```diff
@@ -34,6 +34,8 @@ $grays: map-merge(
 );
 
 $blue:    #275ba9 !default;
+$lightblue: #8cd1ff;
+$grayblue:  #c9e9f6;
```
RyanMulready, line 38: Think it might be best to always include `!default` on variables in this file.

es-bs-base/scss/variables/_grays.scss
```diff
@@ -0,0 +1,18 @@
...
+  gray-100: #fafafa;
+  gray-200: #ececec;
+  gray-300: #dddddd;
+  gray-400: #bababa;
+  gray-500: #999999;
+  gray-600: #7e7e7e;
+  gray-700: #565656;
+  gray-800: #404040;
+  gray-900: #292929;
+  black: #000;
```
RyanMulready, line 17: Shouldn't these reference the variable names and not the hex values?

es-vue-base/src/lib-components/EsReviewModal.vue
```diff
@@ -46,7 +44,9 @@
                                 height="32px"
                                 class="mx-2"
                                 :rating="avgRating" />
-                            <span class="font-size-base font-weight-normal align-self-center">({{ reviews.length }})</span>
+                            <span class="font-size-base font-weight-normal align-self-center"
+                                >({{ reviews.length }})</span
+                            >
```
RyanMulready, line 49: This looks like prettier and eslint are conflicting locally. That shouldn't be happening, prettier itself shouldn't be formatting anything.

RyanMulready (approved):
Comments inline. There does appear to be some unrelated formatting happening in places. Some of the formatting looks correct, but other places it looks odd.

Makefile
tomleo: It breaks if npm packages aren't installed

es-bs-base/scss/_variables.scss
tomleo: So that verticals can override? I guess that makes sense.

es-bs-base/scss/variables/_grays.scss
tomleo: This is for building out the gray color pallet, there was an issue with importing variables. I'm just moving things from extends, but should probably revisit.

es-vue-base/src/lib-components/EsReviewModal.vue
tomleo: Right, that's why I disabled format on save. I'll create a follow-up ticket for this.

PR #7, opened 2022-08-17
Clean up lib import structure

tomleo:
Structure imports in `lib-components` to align with *atomic* naming structures.

The diff is hard to read, I'd just poke around `src/lib-components`, the `index.js` files in particular.


RyanMulready:
So I like the organization of the folders but I'm not sold on the export. I think ideally all the exports come from root or components see below;

```
import { EsButton } from '@energysage/es-vue-base'`
import { EsButton } from '@energysage/es-vue-base/components'`
```

tomleo:
My assumption was that keeping everything top-level (in basically a giant flat object), then tree-shaking seems more strait-forward.

PR #9, opened 2022-09-06
ESDS-53 EsPopover Tests and Link

gigiwargin:
ESDS-53 EsPopover Tests and Link

# Related Jira Tickets

- https://energysage.atlassian.net/browse/ESDS-53

## Changes

- added custom color and updated `font-weight` for links in `EsPopover`
- updated tests for `EsPopover` based on Bootstrap Vue's test [here](https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/components/popover/popover.spec.js), which adds popover to the `document` so that `props` are testable but doesn't add code for popover to `html()`, so I removed the snapshot file
- added `utils` based on Bootstrap Vue's [here] (https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/tests/utils.js)
- made use example more similar to Figma design

### Testing

- Manual

- I wasn't able to achieve all of the same tests as the Bootstrap Vue example, but the tests definitely look a lot better than they did before... do we want me to spend more time trying to get them to work?

### DIFF
--- a/es-design-system/pages/molecules/es-popover.vue
+++ b/es-design-system/pages/molecules/es-popover.vue
@@ -18,10 +18,10 @@
                 <template #title>
                     My Title
                 </template>
-                <p>
+                <p class="mb-0">
                     Install solar panels through this program and get $250 cash back.
                     <a
-                        class="mt-3 d-block"
+                        class="mt-2 d-block"
                         href="https://communitysolar.energysage.com/"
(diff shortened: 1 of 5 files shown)


RyanMulready (approved):
Looks good test locally visual and unit. Impressive work extracting out the test from bootstrap-vue's implementation.

PR #10, opened 2022-09-07
Add additional input types to the form-input component

tomleo:
Was getting:

```
[Vue warn]: Invalid prop: custom validator check failed for prop "type".
```

When adding the form-input to the login vertical. Login form requires `input[type=password]` and `input[type=email]`

## Changes

- Add `email`, `number`, `password`


RyanMulready (approved):
Looks good, what do you think about limiting the types based on what we actually use? That was my suggestion when this was made.

PR #11, opened 2022-09-08
ESDS-69 Popover Color Schemes

gigiwargin:
ESDS-69 Popover Color Schemes

# Related Jira Tickets

- https://energysage.atlassian.net/browse/ESDS-69

## Changes

- Added `customClass` prop to `EsPopover`
- Created custom classes for each color scheme (`es-popover-light` and `es-popover-dark`)

### Testing

- Visual

- The design for the "light" popover in Figma had a gradient as the border color. From what I've read, it's not possible to set a gradient for anything other than `background` in css. I've used `$primary` instead for now. Is there a workaround to use the gradient, or should I ask for approval of a solid color from design?

### How to run locally

- Change the default value of `customClass` in `EsPopover` to view the different color schemes. 
- Change the default value of `placement` to make sure there are no regressions with the arrow/arrow border

## Remaining TODOs

- [x] Add tests for the new prop `EsPopover.spec.js` after ESDS-53 changes are merged in

### DIFF
--- a/es-vue-base/src/lib-components/EsPopover.vue
+++ b/es-vue-base/src/lib-components/EsPopover.vue
@@ -4,7 +4,8 @@
         :show="show"
         :target="target"
         :placement="placement"
-        :triggers="triggers">
+        :triggers="triggers"
+        :custom-class="customClass">
         <template
             v-if="hasTitle"
             #title>
@@ -15,7 +16,7 @@
             <slot name="title" />
             <EsButton
                 variant="link"
-                class="p-0 text-white float-right"
+                class="p-0 float-right"
                 @click="onClose">
                 <XIcon
                     height="20px"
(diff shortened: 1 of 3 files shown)


RyanMulready (commented):
Minor comments inline. I think we should update the [docs](https://github.com/EnergySage/es-ds/blob/8ef0a862994698b6cfcd9719dbcf7e18e22c23bb/es-design-system/pages/molecules/es-popover.vue#L3) and include a `light` display

es-vue-base/src/lib-components/EsPopover.vue
```diff
@@ -111,11 +121,90 @@ export default {
...
+            color: $white;
+        }
+    }
+    .popover-body {
+        .btn, .btn:hover, .btn:active {
+            color: $white;
+        }
+        a {
+        color: #99CBDF;
+        font-weight: $font-weight-bold;
```
gigiwargin, line 206: @RyanMulready I asked design about link styling for both variants and they said that only the `dark` variant has custom styling. The color in ESDS was actually wrong, so I updated it here. I also moved it into the `dark` class definition.

RyanMulready (approved):
Looks good, nice job!

PR #12, opened 2022-09-09
Updates based on design feedback

RyanMulready:
## Changes

- Changed definition of `$pink` to match
- Added missing `!default` and color variable definitions to `_colors.scss`
- Updated badge padding
- Updated `EsButton` documentation and disabled state display
- Fixed some stylelint issues in `EsPopover`
- Updated `EsSlider` to use only necessary color values & changed gradient to be more inline with Figma
- Updated `EsTabs` to default to underline & fixed issue where border color and placement was incorrect
- Updated `EsSupport` to not be reliant upon rep name, also removed image included in bundled package

### Testing

- Visual Manual

-  What do you think about editing css to fit our needs not creating new classes [see here](https://github.com/EnergySage/es-ds/pull/12/files#diff-fdd8380d4d40df79588ad28b8b3e7edb29618a598a695d017fa77f3b511413ddR36) this would be the first time I think we've made a change like this.


es-design-system/pages/molecules/es-support.vue
```diff
@@ -1,17 +1,39 @@
...
+            </template>
+            <template #linkCopy>
+                Link Copy
+            </template>
+        </es-support>
     </div>
 </template>
 <script>
 import { EsSupport } from '@energysage/es-vue-base';
+import benr from '@/assets/imgs/benr.png';
```
tomleo, line 26: I think we should consider using something like https://place-hold.it/300x300 instead of having a png in the project.

tomleo (approved):

es-design-system/pages/molecules/es-support.vue
RyanMulready: Unfortunately not, the image is expected to be a certain ratio and have some transparent area around the person. I understood your argument around not including it in the vue package but whats the aversion to having it in the design system?

es-vue-base/src/lib-components/EsPopover.vue
```diff
@@ -132,58 +132,70 @@ export default {
...
     &.bs-popover-left, &.bs-popover-auto[x-placement^=left] {
         > .arrow {
             &::before {
                 border-left-color: $primary;
             }
+
             &::after {
-                right: 1px;
                 border-left-color: $white;
+                right: 1px;
```
gigiwargin, line 177: Curious what extension I'm missing that didn't catch these 🧐

PR #13, opened 2022-09-09
ESDS-70 Added updated comment to EsReview

gigiwargin:
# Related Jira Tickets

- https://energysage.atlassian.net/browse/ESDS-70

## Changes

- added `updatedComment` prop to `EsReview`
- updated docs and tests with new prop

### Testing

- Manual testing

### How to run locally

- Change the `updated_comment` field in `reviews` in `es-review.vue`. The updated comment should always get shown, if it exists.

### DIFF
--- a/es-vue-base/src/lib-components/EsReview.vue
+++ b/es-vue-base/src/lib-components/EsReview.vue
@@ -126,6 +126,11 @@ export default {
         },
         /**
          * Updated Comment
+         * 
+         * Used in es-cdgm when a user changes 
+         * their original comment this field reflects
+         * the updated comment and comment
+         * retains the original.
          */
         updatedComment: {
             required: false,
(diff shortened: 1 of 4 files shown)


es-vue-base/src/lib-components/EsReview.vue
```diff
@@ -124,6 +124,14 @@ export default {
             required: true,
             type: String,
         },
+        /**
+         * Updated Comment
+         */
```
RyanMulready, line 129: I suspect this is a case where CSM may be the only vertical using this prop. It might be worth putting more explanation into what it is.


```suggestion
        /**
         * Updated Comment
         * 
         * Used in es-cdgm when a user changes 
         * their original comment this field reflects
         * the updated comment.
         */
```

PR #98, opened 2022-09-19
ESDS-74 v-model fixes

RyanMulready:
# Related Jira Tickets

- https://energysage.atlassian.net/browse/ESDS-74

## Changes

- Moved binding of attributes to the component instead of the wrapper for both `EsFormInput` and `EsFormTextarea`

### Testing

- Added new `v-model` test for both input and textarea

### DIFF
--- a/es-design-system/pages/molecules/es-form-input.vue
+++ b/es-design-system/pages/molecules/es-form-input.vue
@@ -4,7 +4,8 @@
             cols="12"
             lg="6">
             <es-form-input
-                id="myId">
+                id="myId"
+                v-model="docInput">
                 <template #label>
                     Account Number
                 </template>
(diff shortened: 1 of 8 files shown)


gigiwargin (commented):
This looks great to me! 🥳 

Only comment is that `eslint` is flagging these errors on `EsFormInput.vue` and `EsFormTextarea.vue` for me:
- `eslint(vue/valid-attribute-name)`
- `eslint(vue/valid-model-definition)`
Are you seeing those?

RyanMulready:
> * valid-attribute-name

I am not and the tests pass in github. Which lines does it think are invalid?

gigiwargin:
>  Which lines does it think are invalid?

It was flagging the first line of the file. Of course now they have disappeared 🙃

gigiwargin (approved):

PR #145, opened 2022-09-30
ESDS-79 used new font-weight-semibold class for EsFormInput/Textarea

ericdouglaspratt:
# Related Jira Tickets

- https://energysage.atlassian.net/browse/ESDS-79

## Changes

- Added `$font-weight-semibold` variable
- Added `font-weight-semibold` utility class
- Switched EsFormInput and EsFormTextarea to use the `font-weight-semibold` utility class on their `<label>` components
- Updated Jest snapshots for EsFormInput and EsFormTextarea
- Added 200 and 300 font weights to the Google Fonts call for the Inter font, since we already have utility classes trying to use those weights
- Added "Font Weight Utility Classes" documentation to the Typography page

### Testing

- Clicked through the `es-design-system` site in desktop Chrome, Firefox, and Safari

- Overall approach, anything I missed

### How to run locally

- `make build` to build `es-bs-base` (where the SCSS changes are) and `es-vue-base` (where the component changes are)
- `make symlink` should hook up those builds to the `es-design-system` Nuxt app
- `npm run dev` in `es-design-system` to run the Nuxt app and see the changes, font weight on the labels should now be 500

## Remaining TODOs

- none

### DIFF
--- a/es-bs-base/scss/_variables.scss
+++ b/es-bs-base/scss/_variables.scss
@@ -362,6 +362,7 @@ $font-weight-lightest:        lighter;
 $font-weight-lighter:         200 !default;
 $font-weight-light:           300 !default;
 $font-weight-normal:          400 !default;
+$font-weight-semibold:        500 !default;
 $font-weight-bold:            600 !default;
 $font-weight-bolder:          800 !default;
 // fusv-disable
--- a/es-bs-base/scss/utilities/_text.scss
+++ b/es-bs-base/scss/utilities/_text.scss
@@ -33,12 +33,13 @@
 
 // Weight and italics
 
-.font-weight-light   { font-weight: $font-weight-light !important; }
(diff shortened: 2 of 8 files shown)


RyanMulready (commented):
Changes look good, nice job. Few minor comments below.

Although we are including `500`; It looks like we aren't including all the required font weights in our google font call. https://github.com/EnergySage/es-ds/blob/3177cb50990e4f891762040469f79d926cd98ddb/es-design-system/nuxt.config.js#L16 Can you add the missing ones?

I assume the above "bug" happened because we aren't including font weight in our typography docs so no one noticed. https://github.com/EnergySage/es-ds/blob/3177cb50990e4f891762040469f79d926cd98ddb/es-design-system/pages/atoms/typography.vue#L4 If you have time it would be awesome to add some minor docs here showing the different font weights.

PR #240, opened 2022-10-25
Add extra-context slot

tomleo:
## Changes

- Add additional extra-context slot to input fields

Need to add an additional line to the login-page on `accounts.energysage` to point users without a password to the OTP flow. Similar to what's currently live on WWW:

Screen Shot 2022-10-25 at 3 16 52 PM

### Testing

- Manual

- Ok if I merge & deploy this?


RyanMulready (commented):
Seems fine to me. Might have called it something besides `extraContext` but not really sure what. `subLabel` maybe but it isn't a label....

Will need to update the snapshots before merging. `npm run test -- -u`

tomleo:
There's also a script `npm run test:update` for that 😉

RyanMulready (approved):

PR #262, opened 2022-11-04
ESDS-93 match header font sizes to figma and make responsive

ericdouglaspratt:
# Related Jira Tickets

- https://energysage.atlassian.net/browse/ESDS-93

## Changes

- Header font size variables now come in two sets: one for mobile, one for desktop
- Header font sizes are now automatically adjusted to the mobile and desktop sizes based on the breakpoint
- Header font sizes now match the [ESDS Figma](https://www.figma.com/file/ILnQorHR2IcGAsIkme9Ctb/ESDS-1.0?node-id=59%3A538)
- H5 and H6 do not appear in the Figma for mobile but are now no larger than H4

### Mobile (before)
Screen Shot 2022-11-04 at 2 34 41 PM

### Mobile (after)
Screen Shot 2022-11-04 at 2 34 55 PM

### Desktop (before)
Screen Shot 2022-11-04 at 2 34 24 PM

### Desktop (after)
Screen Shot 2022-11-04 at 3 10 19 PM

### Testing

- Mobile and desktop testing in Chrome responsive devtools

- Overall approach

### How to run locally

- Build es-bs-base and make symlink
- Run es-design-system site and go to Typography page

## Remaining TODOs

- None

### DIFF
--- a/es-bs-base/scss/_type.scss
+++ b/es-bs-base/scss/_type.scss
@@ -20,6 +20,15 @@ h4, .h4 { @include font-size($h4-font-size); }
 h5, .h5 { @include font-size($h5-font-size); }
 h6, .h6 { @include font-size($h6-font-size); }
 
+@include media-breakpoint-up(lg) {
+  h1, .h1 { @include font-size($h1-font-size-desktop); }
+  h2, .h2 { @include font-size($h2-font-size-desktop); }
+  h3, .h3 { @include font-size($h3-font-size-desktop); }
+  h4, .h4 { @include font-size($h4-font-size-desktop); }
+  h5, .h5 { @include font-size($h5-font-size-desktop); }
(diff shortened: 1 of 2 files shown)


RyanMulready (commented):
Changes make sense, nice job. The reason these values were not "correct" already was to keep them as close to the existing `es-site` and `es-cdgm` existing definitions to reduce regressions. However I think we do want to align with figma definitions and verticals can update as they have time.

With that said this seems like its another breaking change. The regression impacts are fairly significant would you agree? If so I think that means merging this is another major version bump to `0.9.0`?

ericdouglaspratt:
Yes, I agree on the version bump. HPM will need to do a full look at every page when we take this change, as some of our font sizes are dependent on the old sizes.

PR #263, opened 2022-11-05
Add additional validators for password input field

tomleo:
## Changes

- Add additional vuelidate validators for a password field.

### Testing

- Jest tests & manual testing in `es-design-system` project


RyanMulready (approved):
This looks good, FWIW there is an example of how CSM provides it's feedback to the user [here](https://github.com/EnergySage/es-cdgm/blob/master/frontend/components/EsTextInput.vue#L20) would be too hard to add this to `EsFormInput` or create a password specific version of it.

es-vue-base/src/lib-validators/index.js
```diff
@@ -34,3 +34,24 @@ export function vuelidateEmail(emailAddress) {
 export function vuelidatePhone(number) {
     return !helpers.req(number) || /^[+]?[(]?[0-9]{3}[)]?[-\s.]?[0-9]{3}[-\s.]?[0-9]{4,6}$/.test(number);
 }
+
+export const hasNumber = (param) => /[0-9]/.test(param);
+export const hasSpecialCharacter = (param) => /\W|_/.test(param);
```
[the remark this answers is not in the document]
tomleo: Done via a90177c0fdeedf3eb2f97b2e219cc52be2360599

es-vue-base/src/lib-validators/index.js
```diff
@@ -34,3 +34,90 @@ export function vuelidateEmail(emailAddress) {
...
+
+/**
+ * @param { string } param
+ * @returns { boolean } if not required or string contains an lowercase letter
+ */
+export function vuelidateHasLowercaseLetter(param) {
+    return !helpers.req(param) || hasLowercaseLetter(param);
+}
+
+const REQUIRED = 'vuelidateRequired';
```
tomleo, line 94: Add constants here, so as to avoid fingering validator names in validations objects.

tomleo:
There's a fair bit more code I added during integration. The additional `es-form` methods make it easier to list out multiple error messages for complicated validation requirements like a password.

It's probably easiest to look at the reference in `es-design-system` first (the `es-form` example page)

es-design-system/pages/organisms/es-form.vue
```diff
@@ -135,50 +166,70 @@ export default {
...
+                {
+                    name: vuelidateKeys.HAS_UPPERCASE_LETTER,
+                    arguments: '',
+                    description: 'check if the string contains an uppercase character',
+                },
+                {
+                    name: vuelidateKeys.HAS_LOWERCASE_LETTER,
+                    arguments: '',
+                    description: 'check if the string contains a lowercase character',
+                },
```
RyanMulready, line 232: I like that these are individual so we can use them that way but I wonder for this use case we should just have a single `vuelidateKeys.COGNITO_PASSWORD` or something which contains all validators needed to validate a cognito password and their definitions.

es-vue-base/src/lib-validators/index.js
```diff
@@ -34,3 +34,90 @@ export function vuelidateEmail(emailAddress) {
 export function vuelidatePhone(number) {
     return !helpers.req(number) || /^[+]?[(]?[0-9]{3}[)]?[-\s.]?[0-9]{3}[-\s.]?[0-9]{4,6}$/.test(number);
 }
+
+/**
+ * @param { string } param
+ * @returns { boolean } if string contains a number
+ */
+export const hasNumber = (param) => /[0-9]/.test(param);
```
RyanMulready, line 42: I wonder if we should support the amount of "things" required in these validators. This will only tell us if a number is present but lets say we want to check if at least 2 numbers are present.

es-design-system/pages/organisms/es-form.vue
```diff
@@ -258,6 +318,21 @@ export default {
...
+                [vuelidateKeys.HAS_SPECIAL_CHARACTER]: () => 'This field must include a special character',
+                [vuelidateKeys.HAS_UPPERCASE_LETTER]: () => 'This field must include an uppercase letter',
+                [vuelidateKeys.HAS_LOWERCASE_LETTER]: () => 'This field must include a lowercase letter',
+            };
+            const msgFunc = ERROR_MESSAGES[validatorName];
+            if (msgFunc) {
+                return msgFunc();
+            }
+            return '';
+        },
```
RyanMulready, line 335: This feels like it could be part of a mixin.

RyanMulready (commented):
See inline comments. I think there is a benefit here to standardizing what password requirements we will all share. Given `EsForm` is not actually reusable I wonder if creating a specificly reusable `EsFormPassword` component would be more beneficial.

es-vue-base/src/lib-validators/index.js
tomleo: Screen Shot 2022-11-04 at 3 41 10 PM

So this was based upon the password validation from cognito. Their interface seems to allow checking boxes to require the various "things", but there doesn't seem to be hooks for *number of things*.

Is the suggestion to tweak this to be something similar to `minLength` i.e.

```js
hasNumbers = (x=1) => (param) => [...param.matchAll(/[0-9]/g)].length >= x
```

es-design-system/pages/organisms/es-form.vue
tomleo: Yeah, I haven't thought through how this would interface with i18n either. I can pull it out.

RyanMulready (approved):
Optimizations look great, nice job.

PR #316, opened 2022-11-14
Add font size utility classes

RyanMulready:
# Related Jira Tickets

- https://www.figma.com/file/ILnQorHR2IcGAsIkme9Ctb/ESDS-1.0?node-id=59%3A538

## Changes

- Added font size utility classes to match figma definition; based largely on csm's implementation [here](https://github.com/EnergySage/es-cdgm/blob/master/frontend/assets/scss/bs4/_font.scss#L17)
	- Added a few more not in figma following a similar addition pattern

### Testing

- Manual Vis

### DIFF
--- a/es-design-system/pages/atoms/typography.vue
+++ b/es-design-system/pages/atoms/typography.vue
@@ -65,7 +65,8 @@
                     Font Weight Utility Classes
                 </h2>
                 <b-table
-                    :fields="fontWeightFields"
+                    fixed
+                    :fields="fontFields"
                     :items="fontWeightItems"
                     striped>
                     <template #cell(example)="data">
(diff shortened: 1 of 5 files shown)


es-design-system/pages/atoms/typography.vue
```diff
@@ -74,15 +75,34 @@
                 </b-table>
             </b-col>
         </b-row>
+        <b-row class="my-5 border-top pt-5">
+            <b-col>
+                <h2>
+                    Font Size Utility Classes
```
tomleo, line 81: Visual hierarchy is lost on those using screen-readers i.e. utility classes don't convey the semantics of the markup. Ensuring proper hierarchy in headings will also lend to easier navigation & potentially easier parsing by search-engine-crawlers.

Could we add a small note here, maybe something like ["Don't rely only on the appearance of the font (color, shape, font variation, placement, etc.) to convey meaning."](https://a11y-style-guide.com/style-guide/section-general.html#kssref-general-typography)?

RyanMulready: Points taken I will add some documentation. The primary use cases for these classes are not headings or elements meant to convey hierarchy.

tomleo: Totally, they're super useful for building components. Documentation should be enough.

PR #319, opened 2022-11-15
Check if form field doesn't have a related validations def

tomleo:
## Changes

- Add check to `validateState` in the event a form field doesn't include a `validations` definition.

### Testing

- Manual tested in browser & automated test
- To validate the fix, add a form field without related `validations` definitions. Without the check, the page will fail to render throwing an exception.

### DIFF
--- a/es-vue-base/src/lib-mixins/forms.js
+++ b/es-vue-base/src/lib-mixins/forms.js
@@ -66,6 +66,10 @@ export default {
     },
     methods: {
         validateState(name) {
+            // Check that given field has key in validators object
+            if (!this.$v.form[name]) {
+                return null;
+            }
             const { $dirty, $error } = this.$v.form[name];
             return $dirty ? !$error : null;
         },
--- a/es-vue-base/src/lib-mixins/forms.js
+++ b/es-vue-base/src/lib-mixins/forms.js
@@ -67,7 +67,8 @@ export default {
     methods: {
         validateState(name) {
             // Check that given field has key in validators object
-            if (!this.$v.form[name]) {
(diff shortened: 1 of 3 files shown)


es-vue-base/src/lib-mixins/forms.js
```diff
@@ -66,6 +66,11 @@ export default {
     },
     methods: {
         validateState(name) {
+            // Check that given field has key in validators object
+            const fieldHasValidators = this.$v.form && this.$v.form[name];
+            if (!fieldHasValidators) {
+                return null;
+            }
```
RyanMulready, line 73: @ericdouglaspratt pointed out the reliance of the `form` namespace in this method. This might be an opportunity to break that reliance.

RyanMulready (commented):
Looks good, see comment inline about possible further improvement.

tomleo:
> Looks like a fine improvement though I don't really understand the use case.

If the form & validations objects aren't finished, and the mixin is added, the page would crash. This tweak prevents that. So the use case is really developer ergonomics. This change is akin to changing an eslint error to a warning I guess.

I think in general, if using this mixin, we should be adding validators to all the form fields.

es-vue-base/src/lib-mixins/forms.js
tomleo: Is the suggestion to make the mixin functional without the existence of a `form` key in the `data` function?

RyanMulready: That's right. The usage of `validateState(fieldName)` would now become `validateState(form.fieldName)` which is more "correct" imo.

RyanMulready (approved):
Looks good nice changes!

PR #320, opened 2022-11-15
ESDS-97 Update EsVerficationCode to use v-model pattern

RyanMulready:
# Related Jira Tickets

- Removed `v-model` and `$v-on` pattern from child component in favor of `value` and no direct `input`
- `code` is now always treated as an `array` in `EsVeritifcationCode` and as a `string` in its parent
- Greatly simplified pasting behavior; Paste will now always clear existing inputs and start from first input
- Updated documentation to include random and clear code buttons to demonstrate behavior

## Changes

- https://energysage.atlassian.net/browse/ESDS-97

### Testing

- Lots of manual and cross browser testing
- Updated unit tests; still doesn't include paste mocking


es-vue-base/src/lib-components/EsVerificationCode.vue
```diff
@@ -119,89 +144,81 @@ export default {
...
             // Get the paste from clipboard; truncate it if its too long
             this.dataFromPaste = event.clipboardData?.getData('text')
-                .trim().split('').splice(0, this.charCount - prevIndex);
+                .trim().split('').splice(0, this.charCount);
 
-            // Check the data only contains accepted characters
-            const checkPasteValidity = this.checkCodeValidity(this.dataFromPaste);
+            // Reset the code to allow new paste
+            this.code = Array(this.charCount).fill('');
+            this.dataFromPaste.forEach((char, index) => {
```
tomleo, line 172: What do you think about generating the code-array by mapping over an array of empty strings instead of setting the data to an array of empty strings, then updating inline?

Given that `code` is "reactive" is that "safer"? Maybe not given it's writing to a v-dom?

```js
this.code = Array(this.charCount).fill('').map((cur, index) => {
    const pasteVal = this.dataFromPaste[index];
    return pasteVal ? pasteVal : cur;
});
```

es-vue-base/src/lib-components/EsVerificationCode.vue
```diff
@@ -119,89 +144,81 @@ export default {
...
+            // Move focus to the next empty spot or last input
+            this.$refs[`codeInput_${focusIndex}`]?.[0]?.focus();
+
+            this.emitCodeUpdate();
         },
-        checkCodeValidity(code) {
-            if (!code) {
-                return false;
+        // Normalize strings into a valid array
+        stringToArray(value) {
```
tomleo, line 187: Similar to above, what do you think about always returning a fixed length based on `carCount`? Something like:

```js
stringToArray: (value) => Array(this.charCount).fill('').map((cur, index) => value[index] ? value[index] : cur);
```

tomleo (commented):
I like the change, I think this will make it easier to use both es-input and es-verification-code inputs in a form with less differences in API.

Small nits / suggestions inline. I wonder if some logic could be extracted out / shared between `handlePaste`, `stringToArray`, and `valuePropChanged`.

RyanMulready:
Decent amount of changes based on previous feedback. `code` now remains an array throughout for a much cleaner pattern to managing reactivity. The parent component will be in charge of converting it to a string/array based on its needs which feels more correct.

es-vue-base/src/lib-components/EsVerificationCode.vue
```diff
@@ -105,103 +137,86 @@ export default {
...
+            const currentActiveElement = event.target;
+            // Getting the current index handles case where user pastes into an input other than input[0]
+            const curIndex = this.getElementIndex(currentActiveElement);
 
-            // Triggered on a normal input that passes handleKeys bubble
+            // Triggers when a character is added only
             if (!inputType || inputType === 'insertText') {
-                this.checkCodeValidity(this.code);
-
+                // We use $set because replacing by index can remove reactivity
```
tomleo, line 169: 👍

es-vue-base/src/lib-components/EsVerificationCode.vue
```diff
@@ -105,103 +137,86 @@ export default {
...
+            this.code = [...this.emptyArray].map((cur, index) => propValue[index] || '');
 
-            // Test all characters in the code one final time
-            // If a single bad character is found this will short circuit
-            const allValidChars = codeArray.every((num) => {
+            this.emitCodeUpdate(false);
+        },
+        // Emit this.code changes to keep this.value in sync
+        emitCodeUpdate(emitInput = true) {
+            const codeIsValid = this.code.every((num) => {
```
tomleo, line 209: Could this be simplified further into `this.code.every((num) => this.allowedChars.includes(num))`

tomleo (approved):
I haven't tested in downstream nuxt projects, but the changes looks good 👍

PR #321, opened 2022-11-15
POC: Add plugin and pattern to support viewing sources inline

RyanMulready:
## Changes

- Switched `vue` to use `common` which includes runtime compiler
- Added `prism` and its plugins/css to support code highlighting
- Added `plugins/prism` which includes new `CodeBlock` component and supporting methods
- Used `es-badge` documentation as an example of what the plugin can do

- In lieu of real documentation, I often just find myself looking at the source code to determine how to use a component. It seems useful to show the source code for component documentation and the component itself inline instead of expecting users to have the source locally or go to github itself

### How to run locally

- Visit `es-badges` documentation

## Remaining TODOs

- [ ] If pattern seems useful apply to other molecules


tomleo:
This is some impressive hacking!

> In lieu of real documentation, I often just find myself looking at the source code to determine how to use a component. 

Developers & designers I think will/are both consumers of the documentation. I'd argue we do have *real* documentation, it just deviates slightly from most component libraries out there that also have inline code examples.

Personally I feel like having an actual, fully working example is higher fidelity than inline code. This is especially true for things like form validation that require mixins, setting up a data attribute, hooking into input slots for validation, ect.

I like having two vscode windows, and toggling between `es-ds` and `<down-stream-nuxt-repo>`. A lot of the time I can grab most of what's already in an `es-design-system` page.

> It seems useful to show the source code for component documentation and the component itself inline instead of expecting users to have the source locally or go to github itself

I agree inline code-snippets can be useful, and make the documentation look more *legit*. I do however think developers should feel comfortable and be encouraged to dive into the source code. I wonder if something like *view in Github* could be useful?

My only hesitations are that this change will:

- Require maintaining both inline examples and `es-design-system` reference
- `es-design-system` repo will now be slightly more complicated and depart a little more from the structure downstream SPAs will use.

With that being said, this seems handy 👍

RyanMulready:
> I agree inline code-snippets can be useful, and make the documentation look more legit. I do however think developers should feel comfortable and be encouraged to dive into the source code. I wonder if something like view in Github could be useful?

I agree! The functionality to `View Source` is already in the code block, if you hover over it's in the top right. Seems like it might be too hidden there?

tomleo (approved):

ericdouglaspratt (approved):
This is a great and easy to maintain step!

I figure the Documentation Source part will be more useful to consumers of the design system (ES engineers) than the Component Source, so I might advise putting that first.

Longer term, I like the idea of each es-design-system page showcasing several different variants of each component, each with a concise expandable code block attached to it.

And to respond to Tom's point about es-design-system diverging from the SPAs it will be consumed by, I think (a) es-design-system should definitely make very heavy use of the design system itself, but (b) a bit of divergence is inevitable in order for it to become a useful guide for both designers and developers, these code blocks being a good example of that.

PR #385, opened 2022-12-08
Verification bug fixes & color utility classes

RyanMulready:
### ❓ Type of change

- [X] 📖 Documentation (updates to the documentation or readme)
- [X] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

- Updated `EsVerificationCode` to not have `margin` and `min-width` applied; now the responsibility of the parent component
- Added `size` prop to `EsVerificationCode`; defaults to `lg` which is the current size prior to this
- Added a `nextTick` wait to `EsVerificationCode` emits; emitting twice in quick succession results in DOM not updating before second emit
- Updated `EsVerificationCode` documentation to account for changes
- Added `bg-` color classes for all color scales
- Added `text-` color classes for all color scales
- Updated color documentation to reflect new utility classes
- Updated `EsSlider` to not use JS based style definitions and honor SCSS variables instead

### 🥼 Testing

- Updated existing tests and snapshots

### 📝 Checklist

- [X] I have updated the documentation accordingly.
- [X] I have documented testing approach


tomleo (approved):
Looks good :shipit:

PR #393, opened 2022-12-12
Tweaks to form validation

tomleo:
### 🔗 Linked issue

- https://energysage.atlassian.net/wiki/spaces/DS/pages/129925131/2022-11-29+Form+Validation

### ❓ Type of change

### 📚 Description

- Enables error icons in `es-bs-base`
- Add validation mixin helper method for touch validation
- Add examples section to reduce confusion around semantics of `organisms`. Includes documentation around expected validation UX

### 🥼 Testing

- Manual testing in the example page built

### 📝 Checklist

### DIFF
--- a/es-design-system/pages/examples/form-with-validation.vue
+++ b/es-design-system/pages/examples/form-with-validation.vue
@@ -12,10 +12,10 @@
                     <ul>
                         <li>
                             Validation should initially trigger <em>on-blur</em>.<br>
-                            Subsequent validation should trigger on <em>input-change</em>.
+                            Subsequent validation should trigger <em>on-input-change</em>.
                         </li>
                         <li>
-                            Validation for password fields should always trigger on <em>input-change</em>
+                            Validation for password fields should always trigger <em>on-input-change</em>
                         </li>
                     </ul>
                 </p>
@@ -56,7 +56,7 @@
                             <div
                                 v-for="error in formErrors.password"
                                 :key="error">
-                                {{ error }}
+                                {{ getErrorMessage(error) }}
                             </div>
                         </template>
                     </es-form-input>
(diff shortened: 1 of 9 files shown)


es-design-system/pages/examples/form-with-validation.vue
```diff
@@ -0,0 +1,206 @@
...
+                </h2>
+                <p>
+                    <ul>
+                        <li>
+                            Validation should initially trigger <em>on-blur</em>.<br>
+                            Subsequent validation should trigger on <em>input-change</em>.
+                        </li>
+                        <li>
+                            Validation for password fields should always trigger on <em>input-change</em>
+                        </li>
```
ericdouglaspratt, line 19: For consistency, I might suggest a slight copy tweak here to either always have "on" inside the `<em>` or always outside - and I'd suggest removing the hyphenation since, as written, I don't think these refer to specific event names beyond the single words `blur` and `change`.

As an aside - I found on HPM that dropdown selects can't validate on blur, since `<b-dropdown>` doesn't fire a `blur` event - which is mostly fine, since you pretty much only need `required` validation on dropdowns anyway, so once we have an EsFormSelect component, we may need to update this guidance slightly to include a variance for dropdowns.

ericdouglaspratt (changes requested):
Looks great! Added some comments.

Also, I feel like as part of this, we should either update the validation in the EsForm organism page to match, or remove validation from that page entirely and keep that page focused on the purely presentational aspects of inline/not-inline forms. And potentially link to the validation example from there - though admittedly I'm a little fuzzy on our intention for what's an "organism" and what's an "example".

es-design-system/pages/examples/form-with-validation.vue
tomleo: Copy changes via [54b0ac8](https://github.com/EnergySage/es-ds/pull/393/commits/54b0ac8146c3591c8deaceeb5f7536a52c5ac98b)

> once we have an EsFormSelect component, we may need to update this guidance slightly to include a variance for dropdowns

Agreed 👍

ericdouglaspratt (approved):
Looks good! Definitely a great example for people to work from as they use the design system to build forms, a nicer error message experience for users, and some nice work with the icons.

Still want to raise the flag on two items below, but happy to get this merged and we can deal with those separately.
- The coexistence of EsForm and FormWithValidation, with their differing validation patterns, could cause confusion, but happy to discuss organisms vs. examples with the larger group and do a fast follow.
- Minor nit: the copywriter in me still isn't grammatically in love with the hyphens, I'd rather it say "on blur" instead of "on-blur", but people can still get the meaning as is.

tomleo:
@RyanMulready made change from `hasOwn` to `prototype.hasOwnProperty` via [b5c066f](https://github.com/EnergySage/es-ds/pull/393/commits/b5c066f195b63b0399fd0a06e76e39c18c04376b)

PR #455, opened 2023-01-05
Updated EsReview and EsReviewModal to show updates and responses

gigiwargin:
### ❓ Type of change

- [X] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

- Resolves [ESDS-100](https://energysage.atlassian.net/browse/ESDS-100)
- `EsReview` and `EsReviewModal` now account for updates and developer responses to reviews
- Displays correct verified icon, text, comments, and response depending on viewport and if `EsReview` is being used in `EsReviewModal` or not -- all changes based on [this design](figma.com/file/zd3aTvP8hIPJMC7nnpauQd/Review-Display-(CSM-1985)?node-id=315%3A1013&t=FBCGatHgl3gAlwtG-0).

### 🥼 Testing

- Visual/manual testing
- Added to unit tests

### 📝 Checklist

[ESDS-100]: https://energysage.atlassian.net/browse/ESDS-100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


es-design-system/pages/organisms/es-review.vue
```diff
@@ -55,6 +58,7 @@ export default {
                 Mauris vitae tortor mauris. Cras suscipit nibh nec nisi cursus ornare. Maecenas quis turpis sit amet
                 sapien dapibus sollicitudin viverra eu justo. Vivamus posuere metus sit amet purus tempus volutpat.
                 Donec eleifend elit quam.`,
+                response: 'Lorem ipsum',
```
RyanMulready, line 61: Nit: Could we update this to whatever we think a "normal" response length will be?

es-vue-base/src/lib-components/EsReview.vue
```diff
@@ -165,11 +269,33 @@ export default {
...
+         * developer (Number),
+         * response_text (String),
+         * developer_logo (String),
+         * modified (Date)
+         */
+        response: {
+            required: false,
+            type: [Number, String, Date, null],
+            default: null,
+        },
```
RyanMulready, line 290: This should be defined as `type: Object` and then have an expected shape. However, I know it'll be ugly but we should flatten this into 4 different props. Having an prop be an object has caused us lots of headaches in the past.

es-vue-base/src/lib-components/EsReview.vue
```diff
@@ -183,6 +309,14 @@ export default {
...
             return new Date(this.created).toLocaleDateString();
         },
+        updatedDate() {
+            // Displays DD/MM/YYYY
+            return new Date(this.modified).toLocaleDateString();
+        },
+        responseDate() {
+            // Displays DD/MM/YYYY
+            return new Date(this.response.modified).toLocaleDateString();
+        },
```
RyanMulready, line 319: This could be turned into a method to reduce repeating code.

```
<div>
{{ localDate(modified) }}
</div>


localDate(date) {
    return new Date(date).toLocalDateString();
}
```

es-vue-base/src/lib-components/EsReview.vue
```diff
@@ -49,25 +61,108 @@
...
+                <div>
+                    <p class="font-weight-bolder m-0 mb-lg-2">
+                        <span class="d-sm-inline-block">Response from </span>
+                        {{ developerName }}
+                    </p>
+                    <div
+                        class="d-none d-lg-block">
+                        {{ response.response_text }}
+                        <p class="font-size-sm mt-3">
+                            {{ `Responded on ${responseDate}` }}
```
RyanMulready, line 153: TIL you could use this syntax. Below is a bit more "normal".

```suggestion
                           Responded on {{ responseDate }}
```

es-vue-base/src/lib-components/EsReview.vue
```diff
@@ -49,25 +61,108 @@
...
-        <small class="d-none d-lg-flex align-items-center text-gray-800 mt-2">
+        <div class="d-flex flex-nowrap">
+            <small class="d-flex align-items-center text-gray-800 text-nowrap details-holder">
+                <span
+                    class="name-holder d-inline-block text-truncate pr-1"
+                    data-testid="subtext-test">
+                    <span class="d-none d-lg-inline-block">{{ updatedComment && commentLimit ? "Updated by" : "Posted by" }}</span>
+                    {{ reviewerName }}
+                </span>
+                <span class="date-holder">on {{ updatedComment && commentLimit ? updatedDate : formattedDate }}</span>
```
RyanMulready, line 92: Long logic blocks feel bad, this could be a computed property.


```
{{ displayDate }}

displayDate(){
return updatedComment && commentLimit ? updatedDate : formattedDate;
}
```

es-vue-base/src/lib-components/EsReview.vue
```diff
@@ -49,25 +61,108 @@
...
-            </span>
-            <span class="date-holder d-inline-block">on {{ formattedDate }}</span>
-        </small>
-        <small class="d-none d-lg-flex align-items-center text-gray-800 mt-2">
+        <div class="d-flex flex-nowrap">
+            <small class="d-flex align-items-center text-gray-800 text-nowrap details-holder">
+                <span
+                    class="name-holder d-inline-block text-truncate pr-1"
+                    data-testid="subtext-test">
+                    <span class="d-none d-lg-inline-block">{{ updatedComment && commentLimit ? "Updated by" : "Posted by" }}</span>
```
RyanMulready, line 89: This could be a computed property.

RyanMulready (changes requested):
Works well, comments inline. I think the prop change is the only "blocking" one.

es-vue-base/src/lib-components/EsReview.vue
```diff
@@ -165,11 +269,45 @@ export default {
...
             type: [Number, Boolean],
         },
+        /**
+         * Developer Response
+         */
+        response: {
+            required: false,
+            type: [Number, String, Date, null],
+            default: null,
+        },
```
RyanMulready, line 285: Think this should be a string now no?
```suggestion
        response: {
            required: false,
            type: String,
            default: null,
        },
```

es-vue-base/src/lib-components/EsReview.vue
```diff
@@ -190,18 +330,40 @@ export default {
...
+.logo {
+    border-radius: 8px;
+    height: 64px;
+    object-fit: contain;
+    width: 64px;
+    @include media-breakpoint-down(md) {
+        height: 32px;
+        width: 32px;
+    }
+}
```
RyanMulready, line 366: I think there's a cleaner way to accomplish this. For `border-radius` you should be able to use a [rounded](https://github.com/EnergySage/es-ds/blob/7d2bff97201a2d4434a22286971ec0dc58e5058d/es-bs-base/scss/utilities/_borders.scss#L33) class.

For the scaling down of the image take a look at the [fluid](https://bootstrap-vue.org/docs/components/image#component-reference) prop on `b-img`

gigiwargin: I updated the `border-radius` but couldn't get `fluid` to scale the way I expected... 
For testing, I'm using an image that is quite large but can we assume that the uploaded logo will always be the same size?

RyanMulready: It is not safe the assume unfortunately. We suggest `120x120` but allow them to upload whatever size they want.

gigiwargin: Gotcha! The dimensions aren't perfect but I got as close as I could in with my most recent changes in [458f10d](https://github.com/EnergySage/es-ds/pull/455/commits/458f10d0ee55badc409b499bf7e2e643266d8dda)

RyanMulready (approved):
Looks great, appreciate the changes! 🔥

PR #502, opened 2023-01-21
feat: add EsProgress component

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/ESDS-85

### ❓ Type of change

### 📚 Description

- Adds new EsProgress component and accompanying documentation
- Uses Bootstrap Vue's Progress component
- Only changes to default behavior are styling related to match our designs:
    - height is defaulted to 0.3125rem (5px)
    - overall component is not rounded
    - component background is set to a specific color (#ddd)
    - inner progress bar is rounded, to 0.125rem (2px), only when not 100%
- Decided not to include Jest tests for this component since the only functionality/logic we are adding is whether or not the border radius is applied when the value is 100%, which is purely cosmetic and probably not suitable for a Jest test

#### Docs page
Screen Shot 2023-01-21 at 10 16 32 AM

### 🥼 Testing

- Loaded docs page in Chrome, Firefox, Safari and tested animated transition buttons
- Loaded docs page in iPhone/Android emulator with LambdaTest and tested animated transition buttons

### 📝 Checklist

### DIFF
--- a/es-design-system/components/ds-molecules-list.vue
+++ b/es-design-system/components/ds-molecules-list.vue
@@ -56,6 +56,11 @@
             </b-link>
         </li>
         <li>
+            <b-link to="/molecules/es-progress">
+                EsProgress
+            </b-link>
+        </li>
+        <li>
             <b-link to="/molecules/es-rating">
                 EsRating
             </b-link>
--- a/es-design-system/pages/molecules/es-progress.vue
+++ b/es-design-system/pages/molecules/es-progress.vue
@@ -1,0 +1,95 @@
+<template>
+    <div>
+        <h1>
+            Progress
+        </h1>
+        <p class="pb-4">
+            Extended from <b-link href="https://bootstrap-vue.org/docs/components/progress">
+                bootstrap-vue progress
+            </b-link>
+        </p>
+        <h2>
+            Three Steps
+        </h2>
+        <div class="mb-3">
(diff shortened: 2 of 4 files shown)


tomleo (approved):
👍

RyanMulready (commented):
This looks good. I note this will replace `b-progress` but `es-cdgm` and `es-hpm` still have their own version of `WizardProgress` https://github.com/EnergySage/es-cdgm/blob/main/frontend/components/EsWizardProgress.vue https://github.com/EnergySage/es-hpm-pilot/blob/main/frontend/components/EsWizardProgress.vue

Thoughts on if we should extend this component to support their use cases? I know more "wizard" like onboarding flows are coming.

ericdouglaspratt:
The only additional thing the HPM version does is calculate the percentage from the current and total step numbers, which I'd argue should probably be done within a parent wizard component.

Looks like CDGM also has an optional "STEP 1 / 5" label, but the fact that HPM doesn't and that rooftop didn't use a wizard-like layout for 1QPS at all indicates to me that we should have an ESDS guild meeting specific to wizard-type flows to figure out which direction design is going in for one-step-at-a-time flows before we put effort into unifying. The way this progress component is built, it can be used directly on rooftop's 1QPS and CDGM/HPM's wizards with no issue, despite the vast other differences in layout between them.

PR #520, opened 2023-01-25
(feat): Support SSR in es-ds and es-vue-base

RyanMulready:
### ❓ Type of change

- [X] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

- Update `es-design-system` to use `ssr` during development. This will help ensure components are SSR compatible
- Added conditional checks for `$prism` which is client only; ***this is most of the diff***
- Replaced `rollup-terser-plugin`; now deprecated, with `@rollup/plugin-terser`
- Moved `EsSlider` to `client-only`. `vue-slider-component` does not play well with SSR
- Updated `rollup.config.js` for `es-vue-base`
  - Switch to using minified versions by default
  - Added sourcemaps & unminified versions
  - `cjs` is now `main` and `optimizeSSR` is applied to vue components
  - Removed `iife` build in favor of `umd` only

### 🥼 Testing

- Manual, visual and running test builds

### 📝 Checklist

- [X] I have documented testing approach

### DIFF
--- a/es-vue-base/build/rollup.config.js
+++ b/es-vue-base/build/rollup.config.js
@@ -30,6 +30,12 @@ const projectRoot = path.resolve(__dirname, '..');
 
 const baseConfig = {
     input: 'src/entry.js',
+    output: {
+        name: 'EsVueBase',
+        sourcemap: true,
+        exports: 'named',
+        compact: true,
+    },
     plugins: {
         preVue: [
             alias({
@@ -83,7 +89,6 @@ const baseConfig = {
     },
 };
 
-// ESM/UMD/IIFE shared settings: externals
 // Refer to https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
 const external = [
     // list external dependencies, exactly the way it is written in the import statement.
@@ -92,7 +97,6 @@ const external = [
     'bootstrap-vue',
 ];
 
-// UMD/IIFE shared settings: output.globals
 // Refer to https://rollupjs.org/guide/en#output-globals for details
 const globals = {
     // Provide global variable names to replace your external imports
@@ -106,18 +110,14 @@ const buildFormats = [];
 if (!argv.format || argv.format === 'umd') {
     const umdConfig = {
         ...baseConfig,
-        input: 'src/entry.js',
         external,
         output: {
+            ...baseConfig.output,
             file: packageJSON.browser,
-            name: 'EsDsVue',
-            sourcemap: true,
             format: 'umd',
-            exports: 'named',
             globals,
         },
         plugins: [
-            visualizer(), // Outputs bundle info to ./stats.html
             replace(baseConfig.plugins.replace),
             ...baseConfig.plugins.preVue,
             vue(baseConfig.plugins.vue),
@@ -134,6 +134,11 @@ if (!argv.format || argv.format === 'umd') {
                     ],
                 ],
             }),
+            terser({
+                output: {
+                    ecma: 5,
+                },
+            }),
         ],
     };
     buildFormats.push(umdConfig);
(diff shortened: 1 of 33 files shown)


es-vue-base/build/rollup.config.js
```diff
@@ -30,6 +34,11 @@ const projectRoot = path.resolve(__dirname, '..');
 
 const baseConfig = {
     input: 'src/entry.js',
+    output: {
```
tomleo, line 37: 👍

es-vue-base/build/rollup.config.js
```diff
@@ -101,17 +108,73 @@ const globals = {
...
+            ...baseConfig.output,
+            format: 'umd',
+            globals,
+        },
+        plugins: [
+            replace(baseConfig.plugins.replace),
+            ...baseConfig.plugins.preVue,
+            vue(baseConfig.plugins.vue),
+            ...baseConfig.plugins.postVue,
+            babel({
```
tomleo, line 127: Naively I thought transpilation could happen in down-stream projects. Is the problem there that babel doesn't transpile `node_module` packages by default?

es-vue-base/src/lib-components/EsSlider.vue
```diff
@@ -1,36 +1,38 @@
...
+                    {{ label }}
+                </div>
+            </template>
+        </vue-slider>
+    </client-only>
 </template>
 
 <script lang="js">
-import vueSlider from 'vue-slider-component';
+import vueSlider from 'vue-slider-component/dist-css/vue-slider-component.common';
```
tomleo, line 35: 👀

tomleo (approved):
Looks good 👍 

LMK if want me to run through any manual testing.

es-vue-base/src/lib-components/EsSlider.vue
RyanMulready: This thing is a PIA. We should deprecate it.

es-vue-base/build/rollup.config.js
RyanMulready: That's right it ignores everything by default and you have to tell it what packages in `node_modules` you want to transpile.

es-vue-base/tests/__snapshots__/EsSlider.spec.js.snap
```diff
@@ -7,82 +7,90 @@ exports[`EsSlider <EsSlider
...
             ariaLabel="Select a number" /> 1`] = `
-"<div class="es-slider vue-slider vue-slider-ltr" style="padding: 14px 14px; width: auto; height: 4px;" min="100" max="500">
-  <div class="vue-slider-rail">
-    <div class="vue-slider-process" style="height: 100%; top: 0px; left: 0%; width: 50%; transition-property: width,left; transition-duration: 0.5s;"></div>
-    <div class="vue-slider-marks">
-      <div class="vue-slider-mark vue-slider-mark-active" style="height: 100%; width: 4px; left: 0%;">
-        <div class="vue-slider-mark-step vue-slider-mark-step-active"></div>
-        <div class="slider-label vue-slider-mark-label active">
-          100
+"<client-only min="100" max="500">
```
ericdouglaspratt, line 10: Should this `<client-only>` element be output to the HTML?

ericdouglaspratt (commented):
I'm surprised this works! I could have sworn there was an issue at least with with Bootstrap's modal referencing `window` that caused a JS error when I tried switching SSR on for HPM.

I'll defer to Tom for reviewing `rollup.config.js` since I'm not familiar with that, but tested out locally and no issues, so looks good by me.

es-vue-base/tests/__snapshots__/EsSlider.spec.js.snap
RyanMulready: Good catch. This won't happen in the app, it's likely because `jest` isn't aware of what `client-only` is. I'll work on a stub.

PR #550, opened 2023-02-02
add utility classes and examples for loading button

tomleo:
*More details in inline-comments*

### 🔗 Linked issue

- https://energysage.atlassian.net/wiki/spaces/DS/pages/198311937/2023-01-31+Loading+UI+User+feedback+for+long-running+requests+Upcoming

### ❓ Type of change

### 📚 Description

- Added min-width `em` based utility classes (the spinner can go beyond the width of a button if it's width is auto)
- Added width-auto utility classes (previously this was scoped style in a component example)
- Split form validation into two pages, so each are more focused on a particular feature

### 🥼 Testing

- manual (in chrome)

### 📝 Checklist

### DIFF
--- a/es-bs-base/scss/_forms.scss
+++ b/es-bs-base/scss/_forms.scss
@@ -345,3 +345,9 @@ textarea.form-control {
     }
   }
 }
+
+// NOTE: this loading UI is subject to change
+.form-actions__button-spinner {
+  right: 0;
+  top: -1.4em;
+}
--- a/es-bs-base/scss/_forms.scss
+++ b/es-bs-base/scss/_forms.scss
@@ -345,11 +345,3 @@ textarea.form-control {
     }
   }
 }
-
-/* stylelint-disable selector-class-pattern */
-// NOTE: this loading UI is subject to change
-.form-actions__button-spinner {
-  top: -1.4em;
-  right: 0;
-}
-/* stylelint-enable */
--- a/es-bs-base/scss/_forms.scss
+++ b/es-bs-base/scss/_forms.scss
@@ -346,8 +346,10 @@ textarea.form-control {
   }
 }
 
+/* stylelint-disable selector-class-pattern */
 // NOTE: this loading UI is subject to change
 .form-actions__button-spinner {
-  right: 0;
   top: -1.4em;
+  right: 0;
 }
+/* stylelint-enable */
--- a/es-bs-base/scss/utilities/_sizing.scss
+++ b/es-bs-base/scss/utilities/_sizing.scss
@@ -33,7 +33,7 @@
     $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
     // Add min-width (in rem) utilities, especially useful for buttons
     @for $i from 4 through 7 {
-      .min-width#{$infix}-#{$i} { width: #{$i}em !important; }
+      .min-width#{$infix}-#{$i} { width: $i * 1em !important; }
     }
   }
 }
--- a/es-bs-base/scss/utilities/_sizing.scss
+++ b/es-bs-base/scss/utilities/_sizing.scss
@@ -33,7 +33,7 @@
     $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
     // Add min-width (in rem) utilities, especially useful for buttons
     @for $i from 4 through 7 {
-      .min-width#{$infix}-#{$i} { width: $i * 1em !important; }
+      .min-width#{$infix}-#{$i} { min-width: $i * 1em !important; }
     }
   }
 }
--- a/es-bs-base/scss/utilities/_sizing.scss
+++ b/es-bs-base/scss/utilities/_sizing.scss
@@ -18,3 +18,22 @@
 
 .vw-100 { width: 100vw !important; }
 .vh-100 { height: 100vh !important; }
+
+// Width Auto Utils
+@each $breakpoint in map-keys($grid-breakpoints) {
+  @include media-breakpoint-up($breakpoint) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+    .w#{$infix}-auto { width: auto !important; }
+  }
+}
+
+// Min-width EM utils
+@each $breakpoint in map-keys($grid-breakpoints) {
+  @include media-breakpoint-up($breakpoint) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+    // Add min-width (in rem) utilities, especially useful for buttons
+    @for $i from 4 through 7 {
+      .min-width#{$infix}-#{$i} { width: #{$i}em !important; }
+    }
+  }
+}
--- a/es-design-system/pages/examples/form-validation.vue
+++ b/es-design-system/pages/examples/form-validation.vue
@@ -34,9 +34,6 @@
             <b-col
                 cols="12"
                 lg="8">
-                <pre>
-                    isSubmitInProgress: {{ isSubmitInProgress }}
-                </pre>
                 <b-form
                     @submit.stop.prevent="onSubmit">
                     <es-form-msg
--- a/es-design-system/pages/examples/form-validation.vue
+++ b/es-design-system/pages/examples/form-validation.vue
@@ -54,7 +54,6 @@
                         <es-button
                             type="submit"
                             class="w-100 w-lg-auto"
-                            :loading="isSubmitInProgress"
                             :disabled="isSubmitInProgress">
                             <span class="w-100 min-width-6">
                                 <span
--- a/es-design-system/pages/examples/form-validation.vue
+++ b/es-design-system/pages/examples/form-validation.vue
@@ -1,0 +1,128 @@
+<template>
+    <b-container>
+        <b-row>
+            <b-col>
+                <h1>
+                    Form with (form level) Validation
+                </h1>
+                <p>
+                    See
+                    <b-link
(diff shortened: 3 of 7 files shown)


es-bs-base/scss/_forms.scss
```diff
@@ -345,3 +345,9 @@ textarea.form-control {
     }
   }
 }
+
+// NOTE: this loading UI is subject to change
```
tomleo, line 349: Subject to change, via https://energysage.atlassian.net/wiki/spaces/DS/pages/198311937/2023-01-31+Loading+UI+User+feedback+for+long-running+requests+Upcoming#%E2%A4%B4-Decisions

es-bs-base/scss/utilities/_sizing.scss
```diff
@@ -18,3 +18,22 @@
 
 .vw-100 { width: 100vw !important; }
 .vh-100 { height: 100vh !important; }
+
+// Width Auto Utils
+@each $breakpoint in map-keys($grid-breakpoints) {
+  @include media-breakpoint-up($breakpoint) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+    .w#{$infix}-auto { width: auto !important; }
```
tomleo, line 26: This one threw me for a loop as I thought `w-lg-auto` was a utility class. Now it is (pending approval) https://github.com/EnergySage/es-ds/blob/main/es-design-system/pages/organisms/es-form.vue#L342-L346

es-bs-base/scss/utilities/_sizing.scss
```diff
@@ -18,3 +18,22 @@
...
+  }
+}
+
+// Min-width EM utils
+@each $breakpoint in map-keys($grid-breakpoints) {
+  @include media-breakpoint-up($breakpoint) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+    // Add min-width (in rem) utilities, especially useful for buttons
+    @for $i from 4 through 7 {
+      .min-width#{$infix}-#{$i} { width: #{$i}em !important; }
```
tomleo, line 36: Per the [recipe in es-form](https://github.com/EnergySage/es-ds/blob/main/es-design-system/pages/organisms/es-form.vue#L83-L101) I found that a 2em spinner can extend beyond a "width auto" button. A few min-width "in-em" utilities I find really helpful for larger viewports, especially in conjunction with the `w-auto` utility classes.

es-vue-base/src/lib-mixins/forms.js
```diff
@@ -53,10 +53,19 @@ export default {
         * }
         */
         formErrors() {
-            if (!Object.prototype.hasOwnProperty.call(this.$v, 'form')) {
```
tomleo, line 56: Added a bit more error checking so that a page can include the form mixin, but omit a validations section without throwing an error here. In practice this is probably overkill, but the extra defensive-ness here allows form-level validation examples to omit any form-field-level logic which I think is helpful.

es-vue-base/src/lib-mixins/forms.js
```diff
@@ -53,10 +53,19 @@ export default {
...
+            let hasFormAttribute = false;
+            try {
+                hasFormAttribute = Object.prototype.hasOwnProperty.call(this.$v, 'form');
+            } catch (e) {
+                return {};
+            }
```
RyanMulready, line 61: What's the catch here for? We should probably throw an warning instead of swallowing an error.

es-vue-base/src/lib-mixins/forms.js
```diff
@@ -130,7 +139,13 @@ export default {
             this.submitInProgress = false;
         },
         getFields(obj, valueKey = 'obj') {
-            return Object.keys(obj)
+            let objKeys = null;
+            try {
+                objKeys = Object.keys(obj);
+            } catch (e) {
+                return [];
+            }
```
RyanMulready, line 147: `const objKeys = Object.keys(obj) || []`

RyanMulready (approved):
Changes seem fine, few minor nits inline.

es-vue-base/src/lib-mixins/forms.js
tomleo: This is basically catching `Object.prototype.hasOwnProperty.call(undefined)`. I suppose we could throw a warning, but I think this would only trigger when the validations section is omitted from the component object.

es-vue-base/src/lib-mixins/forms.js
tomleo: If `obj` is undefined, then a TypeError will get thrown. These error checks basically allow the mixin to be added to pages without throwing any errors regardless of configuration, whereas before omitting the validators section would cause errors.

es-design-system/pages/examples/form-validation.vue
```diff
@@ -0,0 +1,124 @@
...
+} from '@energysage/es-vue-base';
+
+export default {
+    name: 'EsFormDocs',
+    components: {
+        EsFormMsg,
+        EsFormInput,
+        EsButton,
+    },
+    mixins: [formMixins],
```
tomleo, line 95: With the additional error checking in the `formMixins` the method `startSubmit`, `stopSubmit`, and `showFormError` can be called for form-level validation, without needing to included any client-level validation location i.e. adding the `validators` section required for vuelidate.

es-vue-base/src/lib-mixins/forms.js
tomleo: Here's the context: https://github.com/EnergySage/es-ds/pull/550/files#r1096872668

es-vue-base/src/lib-mixins/forms.js
tomleo: Here's the context: https://github.com/EnergySage/es-ds/pull/550/files#r1096872668

es-design-system/pages/examples/form-validation.vue
```diff
@@ -0,0 +1,124 @@
...
+                        required>
+                        <template #label>
+                            Name
+                        </template>
+                    </es-form-input>
+                    <div class="d-flex flex-grow-1 justify-content-end mt-3">
+                        <es-button
+                            type="submit"
+                            class="w-100 w-lg-auto"
+                            :loading="isSubmitInProgress"
```
ericdouglaspratt, line 57: This `loading` prop currently does nothing, yes? I can't find it in EsButton or in the Bootstrap Vue button docs.

Is it a hint towards how we want to implement this spinner in the future, as part of EsButton? (which I'd agree with)

ericdouglaspratt (approved):
This is a great first stake in the ground towards a common loading pattern!

Also, TIL about just how in-depth the form mixins are with regard to handling error display for you.

There are a number of future enhancements I think we should make for ease of use prior to going org-wide with this pattern (a few listed below), but none of them should block proceeding with this PR:
- IMO this functionality should eventually be a prop of EsButton (e.g. `loading`, `submitting`) which will just take care of it for you
- Most Figma designs I've seen with buttons allow the content to determine the width rather than fixing the width to a certain number of `em`s - I'd likely want to explore EsButton throwing a `visibility: hidden` around button content (so the button's width is still determined by content) when a `loading` prop is true and position-absoluting the spinner in the center of the button.
- We'll want to match the size of the spinner in the future to Huanran's designs (i.e. smaller) and ensure that the spinner also gets small enough to work in the `small` button variant.

This is definitely solid enough to start using in Flash UI and introduce the pattern - nice work!

es-design-system/pages/examples/form-validation.vue
```diff
@@ -0,0 +1,124 @@
...
+                            :disabled="isSubmitInProgress">
+                            <span class="w-100 min-width-6">
+                                <span
+                                    v-if="isSubmitInProgress">
+                                    <b-spinner
+                                        role="status"
+                                        aria-hidden="true"
+                                        label="Loading" />
```
ericdouglaspratt, line 65: It looks like this is how Bootstrap Vue's spinner exposes accessibility - great job including this!

If we move this functionality into EsButton, I can see exposing a loading text slot with "Loading" as a default so users can customize what text to use for describing what's happening - since it looks like spinner component can take `label` as a slot as well.

es-design-system/pages/examples/form-validation.vue
```diff
@@ -0,0 +1,124 @@
...
+                            type="submit"
+                            class="w-100 w-lg-auto"
+                            :loading="isSubmitInProgress"
+                            :disabled="isSubmitInProgress">
+                            <span class="w-100 min-width-6">
+                                <span
+                                    v-if="isSubmitInProgress">
+                                    <b-spinner
+                                        role="status"
+                                        aria-hidden="true"
```
ericdouglaspratt, line 64: Do we need this `aria-hidden`? If we're going to the trouble of including a `label`, we should probably let the browser announce the label to users via the `role="status"` and the `label` rather than hiding it.

es-design-system/pages/examples/form-validation.vue
tomleo: No, I was just copying what was in `pages/organisms/es-form.vue`. Fixed via b121f55c63649a80f17d7971d253fcbbfc88e5b5

es-design-system/pages/examples/form-validation.vue
tomleo: Yup, good catch, fixed via 616bb970e812d7ae4364c2aa25fa970a50b1cc65

PR #611, opened 2023-02-22
feat: add eyebrow heading

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/ESDS-124

### ❓ Type of change

### 📚 Description

- Added eyebrow heading class based on most common patterns seen in designs
- Added eyebrow heading examples to Typography docs page
- Added guidance to Typography page to use sentence case for headings, per marketing style guide
- Updated all headings on the Typography page to be sentence case

#### Eyebrow displayed with other headings
Screen Shot 2023-02-22 at 5 36 59 PM

#### New eyebrow examples
Screen Shot 2023-02-23 at 5 17 43 AM

### 🥼 Testing

- Viewed new heading on docs page

### 📝 Checklist

### DIFF
--- a/es-bs-base/scss/_type.scss
+++ b/es-bs-base/scss/_type.scss
@@ -31,8 +31,8 @@ h6, .h6 { @include font-size($h6-font-size); }
 
 .eyebrow {
   font-weight: $headings-font-weight;
-  letter-spacing: $eyebrow-letter-spacing;
   text-transform: uppercase;
+  letter-spacing: $eyebrow-letter-spacing;
 }
 
 .lead {
--- a/es-bs-base/scss/_type.scss
+++ b/es-bs-base/scss/_type.scss
@@ -29,6 +29,12 @@ h6, .h6 { @include font-size($h6-font-size); }
   h6, .h6 { @include font-size($h6-font-size-desktop); }
 }
 
+.eyebrow {
+  font-weight: $headings-font-weight;
+  letter-spacing: $eyebrow-letter-spacing;
+  text-transform: uppercase;
+}
+
 .lead {
   @include font-size($lead-font-size);
   font-weight: $lead-font-weight;
--- a/es-bs-base/scss/_type.scss
+++ b/es-bs-base/scss/_type.scss
@@ -31,6 +31,7 @@ h6, .h6 { @include font-size($h6-font-size); }
 
 .eyebrow {
   font-weight: $headings-font-weight;
+  color: $headings-color;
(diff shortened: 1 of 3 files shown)


tomleo (changes requested):
Is an "eyebrow" a different visual treatment for headers?

If the intention is to describe the contents below, then semantically it should be a header per [WCAG 2.4.6](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=246#headings-and-labels).

Visual-only treatments are not accessible per [WCAG 1.3.1](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=246#info-and-relationships). Beyond AA adherence, real users seem to value [navigation via headers the most](https://webaim.org/projects/screenreadersurvey9/#finding).

That is all to say, a modified treatment of headers like maybe `<h1><b>Eyebrow</b></h1>` might be ok (or something like `<h1 class="eyebrow">`, but styling a paragraph as a header might have some a11y issues. Unless the intention of eyebrows is to not convey hierarchy (although visually that seems to be the intention).

ericdouglaspratt:
@tomleo That's a great point!

The most common usage in our designs is as a category header directly above another heading.

**Wizard:**
Screen Shot 2023-02-22 at 4 14 14 PM

**Homepage:**
Screen Shot 2023-02-22 at 4 15 18 PM

But it's also being explored as a non-heading type of visual treatment, like to highlight the name of someone who reviewed EnergySage:

**Customer reviews on Homepage:**
Screen Shot 2023-02-22 at 4 16 14 PM

**Highlighted testimonial on Homepage:**
Screen Shot 2023-02-22 at 4 17 12 PM

The use case where it's right above another heading intially had me concerned about using a heading tag for the eyebrow. Screen reader users who are navigating the page via headings will likely associate an `<h3>` eyebrow before an `<h2>` as belonging to whatever `<h2>` appeared before the `<h3>` eyebrow in the DOM, which would be incorrect in this case.

But since semantically, a category **_does_** have a higher abstraction level than the heading it's associated with, we could use a lower-numbered heading tag for the eyebrow (e.g. `<h2>`) and one higher for the heading (e.g. `<h3>`), and then use utility classes to style each however we want.

There is still the non-heading treatment, where a `<p>` is more appropriate, but the most common case should be as a category header, so having the default be a heading tag makes sense.

I've updated the docs to change the `<p>` to an `<h3>` and to add two examples of how to use the eyebrow semantically.

tomleo (approved):
👍

PR #501, opened 2023-01-20
fix: EsButton normalization (hover, active, focus, font size, height, border radius, padding)

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/ESDS-109
- https://energysage.atlassian.net/browse/ESDS-108
- https://energysage.atlassian.net/wiki/spaces/DS/pages/186449921/EsButton+Overhaul

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

- **BREAKING CHANGE:** The `$secondary` color variable is now #292929 (`$gray-900`) instead of #007AAB (`$cyan`). This will cascade through to various styles.
- **BREAKING CHANGE:** Large buttons now exactly match the style of default buttons and have been removed from the example documentation. Their font size is now 16px instead of 20px, border radius is 8px instead of 12px, height is 48px instead of variable based on content, and horizontal padding is 24px instead of 16px.
- **BREAKING CHANGE:** Default and small buttons now have a fixed height (48px and 28px, respectively) that will remain consistent independent of button content, per the design.
- **BREAKING CHANGE:** Default button horizontal padding is now 24px instead of 16px, per the design.
- **DEPRECATION:** The `dark` button variant is now deprecated and is replaced by the `secondary` button variant.
- **DEPRECATION:** The following button variants are now also deprecated: `success`, `info`, `warning`, `danger`, `light`, `mid`, `mid-dark`, `highlight`, and `highlight-dark`.
- Primary and secondary buttons will now vertically center their content by default.
- Primary and secondary buttons have updated hover, active, and focus states, per the design.
- Small button border radius is now 8px instead of 4px, per the design.
- The text color of disabled buttons is now #565656 instead of #292929, per the design.
- The following updates have been made to the EsButton docs page:
   - Guidance on icon sizing is now provided at the top of the page.
   - Button examples are now grouped by variant (e.g. primary, secondary, outline primary, outline secondary).
   - Button examples now show buttons containing text, text with an icon, and icon only.
   - A section has been added to show the Deprecated Buttons, to make it clear for anyone using those variants that they will be removed in a future version.
[...]

PR #613, opened 2023-02-23
feat: new labelSrOnly prop for EsFormInput

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/ESDS-127

### ❓ Type of change

### 📚 Description

- Added a `labelSrOnly` prop to EsFormInput for cases where the form consists of only an input and a submit button within a stylized marketing section like a hero or a CTA banner, and the placeholder text is used as the label for visual users.
- Simplified and broadened the number of examples on the EsFormInput page.

#### Top of EsFormInput page with new basic and required examples
Screen Shot 2023-02-23 at 5 06 30 PM

#### New error and success state examples
Screen Shot 2023-02-23 at 5 06 40 PM

#### New placeholder and hidden label examples
Screen Shot 2023-02-23 at 5 06 49 PM

#### Simplified context message examples
Screen Shot 2023-02-23 at 5 06 55 PM

### 🥼 Testing

- Tested on docs page in Chrome with responsive devtools

### 📝 Checklist

### DIFF
--- a/es-design-system/pages/molecules/es-form-input.vue
+++ b/es-design-system/pages/molecules/es-form-input.vue
@@ -247,10 +247,8 @@ export default {
     data() {
         return {
             form: {
-                docInput: 'Example text',
                 emailCorrect: 'hello@energysage.com',
                 emailWrong: 'hello@energy',
-                example2: 'Example two',
                 firstName: '',
                 password: '',
                 phoneNumber: '',
--- a/es-design-system/pages/molecules/es-form-input.vue
+++ b/es-design-system/pages/molecules/es-form-input.vue
@@ -116,7 +116,7 @@
                     md="6"
                     lg="4">
                     <es-form-input
-                        id="successExample"
+                        id="successMessageExample"
                         v-model="form.emailCorrect"
                         :state="true">
                         <template #label>
--- a/es-design-system/pages/molecules/es-form-input.vue
+++ b/es-design-system/pages/molecules/es-form-input.vue
@@ -16,53 +16,224 @@
                 Sentence case.
             </b-link>
         </p>
-        <b-row class="my-5">
-            <b-col
-                cols="12"
-                lg="6">
-                <es-form-input
-                    id="myId"
-                    v-model="docInput">
-                    <template #label>
-                        Account number
-                    </template>
-                    <template #message>
-                        Please enter your account number.
-                    </template>
-                    <template #errorMessage>
-                        Please enter a valid account number.
-                    </template>
-                    <template #successMessage>
-                        Your account number has been submitted successfully.
-                    </template>
-                </es-form-input>
-            </b-col>
-        </b-row>
-        <b-row class="mt-4">
-            <b-col
-                cols="12"
-                lg="6">
-                <es-form-input
-                    id="myId2"
-                    v-model="example2">
-                    <template #label>
-                        Account Number
-                    </template>
-                    <template #extraContext>
-                        Extra Context Goes Here
-                    </template>
-                    <template #message>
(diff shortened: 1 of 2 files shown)


RyanMulready (commented):
I like the concept here and can think of a few different use cases. I played around with trying to accomplish one of them locally but it ended up illustrating a difference in height between our inputs and buttons that becomes more obvious with a pattern like this.

image


Here it is with a slight adjustment in button height

```
  height: $font-size-base * 3.15; // ensures consistent button height, matching the design, regardless of content
```

image


Thoughts?

ericdouglaspratt:
@RyanMulready Yes, the height difference is a problem I intended to solve separately from this, and I wanted to use the docs example to call out (a) a real-world use case for needing `labelSrOnly`, but (b) this exact issue.

For the moment, on homepage, I'm solving it by using CSS to set the same height as the button, but we'll want to do a PR fairly soon (which I could tackle) to fix. In #501, I deliberately only applied the fixed heights to the button styles, despite us having a pattern of `$input-btn-xxx` variables that make them match, because (a) there was both a default and small height in play and small definitely didn't make sense for inputs, and (b) I wasn't sure yet if inputs needed to be the same height as buttons. Now I am, for this use case.

When we do that, I'd rather not change the button heights, as those are pretty solid and in line with the designs. I'd rather update the input heights to match the default button height, using that `$input-btn-xxx` height variable pattern. But would have to verify that the text size in inputs plus the padding of inputs wouldn't naturally go beyond that height, so I figured a separate PR and ticket for that investigation and fix.

As for whether the button should be separate or appear "joined" to the input (the latter I did initially implement for the stubbed-out homepage components), I've now seen both treatments in the homepage designs (the newer ones have them separated out), so I'd want to check in with design on that before putting effort into joining them.

RyanMulready:
> @RyanMulready Yes, the height difference is a problem I intended to solve separately from this, and I wanted to use the docs example to call out (a) a real-world use case for needing `labelSrOnly`, but (b) this exact issue.
> 
> For the moment, on homepage, I'm solving it by using CSS to set the input to the same height as the button, but we'll want to do a PR fairly soon (which I could tackle) to fix. In #501, I deliberately only applied the fixed heights to the button styles, despite us having a pattern of `$input-btn-xxx` variables that make them match, because (a) there was both a default and small height in play and small definitely didn't make sense for inputs, and (b) I wasn't sure yet if inputs needed to be the same height as buttons. Now I am, for this use case.
> 
> When we do that, I'd rather not change the button heights, as those are pretty solid and in line with the designs. I'd rather update the input heights to match the default button height, using that `$input-btn-xxx` variable pattern. But would have to verify that the text size in inputs plus the padding of inputs wouldn't naturally go beyond that height, so I figured a separate PR and ticket for that investigation and fix. Created [energysage.atlassian.net/browse/ESDS-128](https://energysage.atlassian.net/browse/ESDS-128) to capture this work.

Seems like a reasonable approach to me. Thank you for creating a ticket. 

> As for whether the button should be separate or appear "joined" to the input (the latter I did initially implement for the stubbed-out homepage components), I've now seen both treatments in the homepage designs (the newer ones have them separated out), so I'd want to check in with design on that before putting effort into joining them.

I suppose offering both in documentation isn't the best idea even if both patterns exist in the wild. I think asking for clarification from design on which should be the documented pattern is right.

ericdouglaspratt:
@RyanMulready Just asked in the ESDS core team channel and both Huanran and Maria confirmed the separated input and button are the finalized version, and the joined version was just an exploration, so this should be good to go.

RyanMulready (approved):

PR #625, opened 2023-02-28
fix: issues with link variant of EsButton

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/ESDS-118

### ❓ Type of change

### 📚 Description

- Added an `inline` prop for EsButton, only for use with the `link` variant, that removes the padding, height, border, etc. to make it align perfectly with text it may be next to, so it looks like a regular link
- Added link variant examples to EsButton docs page
- Added a prop table to EsButton docs page
- Fixes a small issue with EsCollapse introduced in #501 where the expand/collapse button was slightly shorter than it should be in cases where it contained large text (it now aligns horizontally with the text content beneath it even better than it did prior to the introduction of the issue)
- Fixes an issue with EsViewMore introduced in #501 where the "Show Less" link was too tall (it now aligns with the text even better than it did prior to the introduction of the issue)

#### Link button examples on EsButton docs page
Screen Shot 2023-02-28 at 2 35 05 PM

#### New props table on EsButton docs page
Screen Shot 2023-02-28 at 2 36 03 PM

#### EsViewMore button link now aligns with text
Screen Shot 2023-02-28 at 2 33 41 PM

### 🥼 Testing

- Tested in Chrome responsive dev tools, Firefox, and Safari
- Tested in iPhone/Android emulator with LambdaTest

### 📝 Checklist

### DIFF
--- a/es-vue-base/src/lib-components/EsButton.vue
+++ b/es-vue-base/src/lib-components/EsButton.vue
@@ -1,5 +1,6 @@
 <template>
     <b-button
+        :class="{ 'inline': inline }"
         :variant="computedVariant"
         :size="size"
         v-bind="$attrs"
@@ -45,6 +46,14 @@ export default {
             validator: (val) => ['lg', 'md', 'sm'].includes(val),
             default: 'md',
         },
+        /**
+         * Only works for 'link' variant buttons.
+         * Removes the fixed height and padding so the button can fit nicely within a block of text.
+         */
+        inline: {
+            type: Boolean,
+            default: false,
+        },
     },
     computed: {
         /**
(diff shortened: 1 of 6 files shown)


es-vue-base/src/lib-components/EsButton.vue
```diff
@@ -60,3 +69,15 @@ export default {
     },
 };
 </script>
+
+<style lang="scss" scoped>
+.btn-link.inline {
```
tomleo, line 74: I wonder if this should go in `es-bs-base`, but we can always extract it out later

tomleo (approved):
👍

es-vue-base/src/lib-components/EsButton.vue
ericdouglaspratt: Possibly? I guess you could manually apply the `inline` and `btn-link` classes to a `<button>` on one of our BS2->BS4 pages - would have to confirm that's a need.

PR #769, opened 2023-04-17
feat: add EsNavBar global navigation component

mpleroux:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-390

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

- This `navigation-redesign-sandbox` branch includes all navbar work completed to date
- Created `EsNavbar` component and added it to Organisms page
- Removed test `GlobalNavSkeleton` version of code from Molecules page
- Moved SVGs for dropdown menu CTAs into separate components and stored them in a new directory `es-vue-base/src/lib-assets` for now
    - Is that location appropriate?
    - Should they be components or regular SVG files?
- Couldn't find a way to load JSON data from external file instead of storing it in `data()`
    - Is that a good idea?
- The HTML and SCSS could use some refactoring after next week's demo. There are some open tickets to address that in the Navigation Redesign epic.
- So far the navbar has been built without using many Design System components, BootstrapVue components, or Bootstrap helper classes. It's uncertain how much of that would work in the static "ripped" version to be used in older Bootstrap 4 pages.
- `rip-the-nav` fails when parsing the script because comments no longer appear in the minified JS

### 🥼 Testing

- Tested locally
- Haven't created new Jest test for `EsNavBar` yet

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.
- [ ] I have documented testing approach

PR #770, opened 2023-04-17
feat: Build footer

jsieving:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-376

### ❓ Type of change

### 📚 Description

- Added `EsFooter` molecule to the design system
- I tried to avoid using custom styles, so there are a couple places I compromised on the designs, e.g. having a 3rem margin off by 4px
- Smaller SVGs are hard-coded in as specified in the ticket, but at some point we may want to extract them as in https://github.com/EnergySage/es-ds/pull/769.
- One SVG was unnecessarily detailed and made up about 3/4 of the compiled HTML component, so I replaced it with the PNG used in the existing footer.
- Some URLs still need to be added, as the pages need to be verified or don't exist yet.

### 🥼 Testing
- Manual testing on demo server and on local dev environment of es-site
- There is a basic Jest test, but I'm not sure if it needs further testing since it's a static component

### 📝 Checklist


jimlind:
Incredibly quick turn around and following your intuition to fill in the gaps between mocked sizes makes the responsive footer work great.

In the mock the social icons are always left aligned but in your version they go full aligned.
Both are valid choices but we should follow the mocks here.

Screen Shot 2023-04-18 at 9 53 38 AM
Screen Shot 2023-04-18 at 9 53 44 AM

es-vue-base/src/lib-components/EsFooter.vue
```diff
@@ -0,0 +1,158 @@
...
... 2.5-5 0-2 .2-5.1-2.6-5.1ZM114.5 19.2c0 3.7 1 4.5 2.7 4.5 1.5 0 2.3-1.2 2.3-2.6h2.4c0 3-1.8 4.4-4.6 4.4-2.9 0-5.2-.9-5.2-6.8 0-4 .5-7.2 5.2-7.2 3.8 0 4.7 2 4.7 6.5v1.2h-7.5Zm5.2-1.7c0-3.9-1-4.4-2.7-4.4-1.3 0-2.5.7-2.5 4.4h5.2Z"/><path fill="#616161" d="M154 12c-.9-1.7-2.3-3.5-4-5.4-2.8-3.2-5.6-5.6-5.7-5.7a2.2 2.2 0 0 0-2.8 0 62.2 62.2 0 0 0-9.8 11c-1.4 2.4-2 4.4-2 6.2 0 7 5.3 12.6 12 13.2v3.3c0 .7.6 1.2 1.2 1.2s1-.5 1-1.2v-3.3c6.8-.6 12.2-6.3 12.2-13.2 0-1.8-.7-3.8-2-6.1ZM143.5 29h-1a11 11 0 0 1-10.5-11c0-6 11-15.5 11-15.5s11 9.4 11 15.5a11 11 0 0 1-10.5 11Z"/><path fill="#616161" d="M147.3 15.9a1 1 0 0 0-1-.4l-1.8.3.4-7.7a1 1 0 0 0-1.8-.5l-4.7 11.1a1 1 0 0 0 0 1c.2.4.6.6 1 .5l1.8-.3L140 26a1 1 0 0 0 1 1h.1c.3 0 .5-.3.7-.5l5.5-9.7a1 1 0 0 0 0-1m-1.2.8-4.7 8.3 1-6.5-2.9.5 4.2-9.7-.4 7.9 2.8-.5ZM158.6 22.6a1.4 1.4 0 0 0-1.5 1.5c0 .8.7 1.4 1.5 1.4.7 0 1.4-.6 1.4-1.4 0-.8-.7-1.5-1.4-1.5Zm1.1 1.5c0 .6-.5 1.1-1.1 1.1-.7 0-1.2-.5-1.2-1.1 0-.7.5-1.2 1.2-1.2.6 0 1.1.5 1.1 1.2Z"/><path fill="#616161" d="m159 23.8-.1-.3-.4-.1h-.4v1.3h.2v-.5h.2l.4.5h.2l-.4-.6c.2 0 .3-.1.3-.3Zm-.5.2h-.2v-.4h.4v.4h-.2Z"/></g><defs><clipPath id="a"><path fill="#fff" d="M0 .4h160v35.3H0z"/></clipPath></defs></svg>
```
mpleroux, line 31: I'm concerned about the IDs and URLs in the SVGs.

It's a pain to hand-edit SVG code, but since these footer SVGs will be included in every page we should [prefix them](https://github.com/EnergySage/es-site/blob/develop/docs/frontend/svg.md#process) (with something like `footer-`?) to avoid conflicting with other SVGs.

I think it's inevitable that some other exported SVG will use the same `#a`. What do you think?

mpleroux (commented):
I assume a component more complex than a typical molecule should be moved to `organisms` eventually.

Should the "Make an impact" section at the top of the Figma designs be part of this ticket?

es-vue-base/src/lib-components/EsFooter.vue
RyanMulready, line 31: Rule of thumb, svg's should only be inlined if they are iconography. If they are in fact real images they are better served as an normal asset using `b-img src="mySvg.svg"`

es-vue-base/src/lib-components/EsFooter.vue
```diff
@@ -0,0 +1,158 @@
...
+                    </a>
+                    <a href="https://twitter.com/energysage">
```
RyanMulready, line 50: Could these other links/copy also be part of the `data()` response?

es-vue-base/src/lib-components/EsFooter.vue
jimlind: This specific SVG is the EnergySage typeface+leaf which doesn't exist in the es-ds yet but will after that gets merged. I was thinking it would be a non-blocking fast follow to replace with the actual logo but could prioritize it if that makes sense: https://energysage.atlassian.net/browse/CED-413

jimlind: This specific SVG is the EnergySage typeface+leaf which doesn't exist in the es-ds yet but will after that gets merged. I was thinking it would be a non-blocking fast follow to replace with the actual logo but could prioritize it if that makes sense: https://energysage.atlassian.net/browse/CED-413

es-vue-base/src/lib-components/EsFooter.vue
```diff
@@ -0,0 +1,158 @@
...
+        </div>
+        <hr class="border-top border-dark m-0">
+        <!-- Trademark Info -->
+        <p class="mt-150 mt-lg-200">
+            ENERGYSAGE is a registered trademark and the EnergySage logo is a trademark of EnergySage, Inc.
+            Other trademarks are the property of either EnergySage, Inc. or our licensors
+            and are used with permission.
+        </p>
+        <p class="mb-200">
+            © Copyright 2009-2022 EnergySage, Inc. All rights reserved.
```
RyanMulready, line 76:
```suggestion
            © Copyright 2009-{{ new Date().getFullYear() }} EnergySage, Inc. All rights reserved.
```

es-vue-base/tests/EsFooter.spec.js
```diff
@@ -0,0 +1,18 @@
+import { mount } from '@vue/test-utils';
+import EsFooter from '@/src/lib-components/EsFooter.vue';
+import jestVue from '@/tests/jest.vue.config';
+
+describe('EsFormMsg', () => {
+    // Basic test; should exist for most components
+    test('<EsFormMsg />', async () => {
```
RyanMulready, line 7:
```suggestion
describe('EsFooter', () => {
    // Basic test; should exist for most components
    test('<EsFooter />', async () => {
```

es-vue-base/tests/__snapshots__/EsFooter.spec.js.snap
```diff
@@ -0,0 +1,200 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`EsFormMsg <EsFormMsg /> 1`] = `
```
RyanMulready, line 3:
```suggestion
exports[`EsFormMsg <EsFooter /> 1`] = `
```

es-vue-base/src/lib-components/EsFooter.vue
jimlind: Up to Jane if this is a quick fix. These links should be changing even less than the ones in the data method so building that logic feels very optional.

jsieving: If we factored out the social icons I think this would make sense, but I'm inclined to leave it as is. _If_ these change, it feels like extracting just the names and URLs will make it more likely that we end up with mismatched links & icons.

jsieving:
> Both are valid choices but we should follow the mocks here.

Bye-bye, lack of custom styles 🥲 

> Should the "Make an impact" section at the top of the Figma designs be part of this ticket?

Apparently it is... I hallucinated a comment that I was supposed to ignore it. Will do.

es-vue-base/src/lib-components/EsFooter.vue
RyanMulready: I will note the file does exist in the [navigation](https://github.com/EnergySage/es-ds/pull/769/files#diff-4bd3f38780247990e68745676ee468f510b020477d3fedc08a1ce71f573e2c92) branch. The DOE logo should get similar treatment.

jsieving: We're using the PNG from the current site footer for the DOE logo because the SVG for that logo is ridiculously detailed/oversized and still a little broken. I can copy over the ES logo from the other branch, but it seems simpler to do that after one of these PRs is merged.

jsieving:
I think this is ready for final review.

RyanMulready (approved):
This looks great. 🔥  I appreciate the quick turn around on suggestions/feedback. ❤️

tomleo (changes requested):
The feature looks great, however I might suggest upgrading `es-footer` to an *organism* vs. a *molecule*.

I'm also not sure if the footer should be packaged in `es-vue-base`, I think having it in the `es-design-system` should be enough.

Here's what I was thinking https://github.com/EnergySage/es-ds/pull/779

tomleo (approved):
👍

es-vue-base/src/lib-components/EsFooter.vue
```diff
@@ -0,0 +1,156 @@
...
+        </div>
+    </div>
+</template>
+
+<script lang="js">
+export default {
+    name: 'EsFooter',
+    data() {
+        return {
+            columns: {
```
ericdouglaspratt, line 110: Optional: In cases where we care about preserving the order of a set of items in JavaScript, I'd generally still recommend using an array, even though in most modern browser implementations, the order of chronological definition will actually be preserved and this will work. So not a necessary change but I did change the nav data objects to be arrays just for extra precaution.
https://www.stefanjudis.com/today-i-learned/property-order-is-predictable-in-javascript-objects-since-es2015/

es-design-system/pages/atoms/icons.vue
```diff
@@ -376,6 +376,37 @@
             </li>
         </ul>
 
+        <h2>
+            Social Icons
+        </h2>
+        <p>
+            Default size for social icons is 31px wide by 32px tall.
```
ericdouglaspratt, line 383: Nit: I'd love if we could eventually get 32px by 32px assets from design so these can match the other icons in being an exact square, but that can be done later.

es-design-system/pages/molecules/es-footer.vue
```diff
@@ -0,0 +1,38 @@
+<!-- eslint-disable vuejs-accessibility/no-static-element-interactions -->
+<template>
+    <div>
+        <h1>
+            EsFooter component
+        </h1>
+        <es-footer />
```
ericdouglaspratt, line 7: Optional: Would be great if the docs page used a layout where the footer could go full-width so we can give it a proper layout test, but we can also do that later.

ericdouglaspratt (approved):
Looks good! A few small things we may want to do in the future, but nothing that needs to block merging this in now.

PR #779, opened 2023-04-20
CED-376 change footer to be an organism

tomleo:
Looking at the [Atomic Design Book](https://atomicdesign.bradfrost.com/chapter-2/#organisms) I think the footer is better classified as an organism.


tomleo:
Going to table this for now.

PR #789, opened 2023-04-25
feat: add reviews.io component

nathanielwarner:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-408

### ❓ Type of change

### 📚 Description

Adds the reviews.io component that is currently defined in `es-cms-pages`. It will be removed from there after this change is published.

In the course of migrating the component, I noticed a bug, where navigating away from the page leaves behind the text that was in the reviews.io review. It appears that this text makes up the modals that are normally hidden, so it's not visible on the page itself. When navigating away, the styles disappear so it is visible. To fix this, I added the hook `beforeRouteLeave` to remove this left-behind text upon leaving the page.

### 🥼 Testing

Manual testing

### 📝 Checklist


ericdouglaspratt (changes requested):
Looks great overall, thanks for putting this in!

One thing - I think we should align with the naming convention of the other molecules/organisms in the design system by prefixing with "Es".

nathanielwarner:
> Looks great overall, thanks for putting this in!
> 
> One thing - I think we should align with the naming convention of the other molecules/organisms in the design system by prefixing with "Es".

All set!

ericdouglaspratt (approved):

lgeggleston (changes requested):
Hi! Code looks good, thanks for catching that bug & adding the `beforeRouteLeave` as well!

The component itself needs to be in `es-vue-base/src/lib-components` though I think; `es-design-system` is just the [documentation pages](https://github.com/EnergySage/es-ds#the-parts) that display an example of that component in use.
[This informal doc](https://energysage.atlassian.net/wiki/spaces/CPod/pages/249102350/Using+the+design+system+notes) is helpful too for testing with that pattern & rebuilding `es-vue-base` to see changes.

ericdouglaspratt:
Lia's absolutely right, it needs to be in `es-vue-base` and then imported into `es-design-system` to display on the docs page - apologies for the miss - it's been a hectic day!

nathanielwarner:
> es-design-system is just the [documentation pages](https://github.com/EnergySage/es-ds#the-parts) that display an example of that component in use

That... actually makes a lot of sense. My apologies, as I've never worked in this repo before!

lgeggleston (approved):
Looks good! Tested on multiple browsers & viewports :ship:

PR #829, opened 2023-05-04
feat: EsFooter/EsNavBar link domains are now customizable

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-509

### ❓ Type of change

### 📚 Description

- `es-vue-base` now exports three new utility functions: `getEsFooterContent`, `getEsNavBarAccountContent`, and `getEsNavBarAccountContent` - these can take optional parameters of the various EnergySage domain names, but will also take the same named variables from `process.env` if present
- EsFooter now takes `content` as a prop instead of defining its content in its `data()` function
- EsNavBar now takes `accountContent` and `globalContent` as props instead of importing directly from a JS file
- Consuming applications of EsFooter and EsNavBar can now provide the appropriate domain names for links based on environment (e.g. local, dev) by passing them into those functions - or they can pass in any content they want
- Fixed five links that were `http` to `https`
- Moved remaining hardcoded text in EsNavBar out to the prop data

### 🥼 Testing

- Tested in Chrome on docs page, ensured footer and nav bar render properly
- Tested passing in `https://www.google.com` as the first argument to the content functions and verified that the domain of all appropriate links updated accordingly
- Tested setting up `.env` and the `nuxt.config.js` pass-through of env variables in `es-design-system` and verified that the `process.env.ES_DOMAIN` variable was used correctly

### 📝 Checklist


es-design-system/layouts/nav.vue
```diff
@@ -1,6 +1,8 @@
 <template>
     <div>
-        <es-nav-bar />
+        <es-nav-bar
+            :account-content="accountContent"
+            :global-content="globalContent" />
```
RyanMulready, line 5: Might be out of scope but feels like `fetch('/api/account-first-name/',` could also be a prop here and give us an easier path to changing the logic.

```javascript

props: {
  isUserLoggedIn: {
    type: Function,
    required: false,
    default: () => false,
  }
}
```

es-vue-base/src/lib-components/EsFooter.vue
```diff
@@ -119,66 +119,11 @@ export default {
...
-                { text: 'Privacy policy', url: 'https://www.energysage.com/privacy-policy/' },
-                { text: 'Mobile terms of use', url: 'https://www.energysage.com/mobile-terms-of-use/' },
-                { text: 'Non-discrimination policy', url: 'https://www.energysage.com/nondiscrimination/' },
-            ],
-        };
+    props: {
+        content: {
+            type: Object,
+            required: true,
+        },
```
RyanMulready, line 126: Be super nice if we had typescript here 🤔

es-vue-base/src/lib-utils/footer-content.js
```diff
@@ -0,0 +1,65 @@
+export default (
+    ES_DOMAIN = 'https://www.energysage.com',
+    NEWS_DOMAIN = 'https://news.energysage.com',
+    COMMUNITY_SOLAR_DOMAIN = 'https://communitysolar.energysage.com',
+    HEAT_PUMPS_DOMAIN = 'https://heatpumps.energysage.com',
+    HELP_DOMAIN = 'https://help.energysage.com',
```
RyanMulready, line 6: I wonder if we add `process.env` calls in here it would allow us to override the default variable at build time.

```javascript
    ES_DOMAIN = process.env.ES_DOMAIN || 'https://www.energysage.com',
```

es-design-system/layouts/nav.vue
ericdouglaspratt: I think that is out of scope for this ticket. I also don't know a way of implementing that without breaking the ripped `es-site` version. Some possibilities:
1. Always make the hardcoded fetch API call and ignore the result if some prop is passed in (but then the API call would be made unnecessarily on HPM/CSM).
2. Have some second boolean prop defaulted to true that represents whether to make the hardcoded fetch API call or not, and in the non-Vue case, somehow do a whole bunch of existence checking to get around the fact that the `this.[propName]` reference will break in vanilla JS? Not sure.

Either way I think this should wait until the ticket to make the nav HPM/CSM compatible.

RyanMulready: With flash now live this is a lot easier; https://energysage.atlassian.net/browse/CED-511

es-vue-base/src/lib-utils/footer-content.js
ericdouglaspratt: Good suggestion, that worked! Tested it with a `.env` file and `nuxt.config.js` pass-through in `es-design-system` and it changed the domain nicely - definitely simplifies a step in consuming applications.

RyanMulready (approved):
Looks good to me. Minor suggestion; in this code base it's `ES_DOMAIN` in CMS code bases it's `ENERGYSAGE_WWW_DOMAIN` one of them should change, no strong opinion on which.

ericdouglaspratt:
Yeah I was planning to update `es-cms-pages` to `ES_DOMAIN`, since the `WWW` qualifier seems unnecessary for the future since everything will eventually be on that domain.

PR #852, opened 2023-05-12
feat: EsFooter and EsNavBar links are now SPA-compatible

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-509

### ❓ Type of change

### 📚 Description

- All links in `EsFooter` and `EsNavBar` now use a new `EsNavBarLink` component
- `EsNavBarLink` by default renders a `<b-link>` tag - if the `href` it's given as a prop starts with a `'/'`, it passes it to the `<b-link>` as a `to` prop, otherwise uses an `href` prop - this will allow consuming Nuxt applications (if they pass in a customized set of links to `EsFooter` or `EsNavBar`, some of which are root-relative rather than absolute) to navigate between Nuxt pages like a single-page application without full page reload

### 🥼 Testing

- First tested this by cloning all EsNavBar code into a Nuxt app and confirming that SPA behavior occurred on root-relative URLs and normal link behavior occurred on other URLs
- Confirmed that a simpler approach of hardcoding all links to be `<b-link>` and to use the `to` prop would **not** work because it rendered absolute URLs in front of the current page's URL like `http://localhost:8500/organisms/https://www.energysage.com` which will not work.
- Tested all types of links in the footer and nav on the ESDS local docs site to confirm they all still work as expected

### 📝 Checklist

### DIFF
--- a/es-vue-base/src/lib-components/EsNavBarLink.vue
+++ b/es-vue-base/src/lib-components/EsNavBarLink.vue
@@ -1,0 +1,31 @@
+<template>
+    <component
+        :is="tag"
+        :href="isRootRelative ? '' : href"
+        :to="isRootRelative ? href : ''"
+        v-bind="$attrs"
+        v-on="$listeners">
+        <slot />
+    </component>
+</template>
+
+<script lang="js">
+export default {
+    name: 'EsNavBarLink',
+    props: {
+        href: {
+            type: String,
+            default: '',
+        },
+        tag: {
+            type: String,
+            default: 'b-link',
+        },
+    },
+    computed: {
+        isRootRelative() {
+            return this.href.startsWith('/');
+        },
+    },
+};
+</script>
--- a/es-vue-base/src/lib-components/EsFooter.vue
+++ b/es-vue-base/src/lib-components/EsFooter.vue
@@ -32,12 +32,12 @@
                                     v-for="link in column.links"
                                     :key="link.text"
                                     class="mt-25 mb-50">
-                                    <a
+                                    <es-nav-bar-link
                                         :href="link.url"
                                         :target="link.newTab ? '_blank' : null"
                                         class="text-reset">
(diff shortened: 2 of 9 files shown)


es-vue-base/src/lib-components/EsNavBarLink.vue
```diff
@@ -0,0 +1,31 @@
+<template>
```
tomleo, line 1: Is there anything unique to the nav-bar? I guess I wonder if relative links defaulting to `:to` vs. absolute defaulting to `:href` could be used beyond the nav i.e. could this be an `EsLink` component?

tomleo (approved):
Overall looks good, one question inline.

es-vue-base/src/lib-components/EsNavBarLink.vue
ericdouglaspratt: I could see an `EsLink` component being used internally to ESDS in cases where an exported component contains links, the behavior of which, downstream applications may want to modify. There probably aren't too many cases of this in ESDS. Nav bar and footer are a bit of a special case that won't live within ESDS for too much longer anyway, but definitely something to keep in mind for the future if we ever find another case within ESDS.

It's also a very opinionated behavior that may not work in all cases - there are probably cases where you'd want to allow root-relative URLs to be normal `href` attributes, if an application was only on one domain, wasn't an SPA, and that's how it wanted to specify its URLs. But that does seem unlikely in our new tech stack, at least for the moment.

jsieving (approved):
I checked this out locally and it looks good to me. Is it automatically adding `noopener` to different-domain links and `target="_self"` to same-domain ones?

ericdouglaspratt:
@jsieving Yes that's an artifact of using `<b-link>` rather than just `<a>` - since that component follows a number of standard best practices that we may not have been doing before.

PR #912, opened 2023-06-01
fix: a11y bugs

RyanMulready:
### ❓ Type of change

### 📚 Description

- Fixed issue in `EsRating` where `0` was not setting correct aria attribute
- Fixed issue in the navbar where `li` existed outside of `ul`
- Fixed issue in the navbar where `ul` existed with no `li` in it
- Fixed issue with unit tests and unknown icon components

### 🥼 Testing

- Updated snapshots

### 📝 Checklist


es-vue-base/src/lib-components/EsNavBarTopicMenu.vue
```diff
@@ -94,7 +94,7 @@
                     </es-nav-bar-link>
                     <div
                         v-else
-                        class="nav-item col-12">
+                        class="nav-item pl-3">
```
ericdouglaspratt, line 97: This should probably be `pl-100` to match the new spacing naming convention.

ericdouglaspratt (changes requested):
I'm seeing a few styling regressions as a result of this. We probably also want to wait for Mike to get back so he can test out the nav and footer changes on es-site.

On the mobile EsNavBar, if you go into the "EV chargers" menu (or any of the product menus), the divider line isn't going all the way across and some of the items aren't wrapping and are rendering next to each other. I'm also seeing a horizontal scroll bar in the menus where items appear next to each other.
Screen Shot 2023-06-01 at 10 29 05 AM
Screen Shot 2023-06-01 at 10 33 09 AM

RyanMulready:
> I'm seeing a few styling regressions as a result of this. We probably also want to wait for Mike to get back so he can test out the nav and footer changes on es-site.
> 
> On the mobile EsNavBar, if you go into the "EV chargers" menu (or any of the product menus), the divider line isn't going all the way across and some of the items aren't wrapping and are rendering next to each other. I'm also seeing a horizontal scroll bar in the menus where items appear next to each other. Screen Shot 2023-06-01 at 10 29 05 AM Screen Shot 2023-06-01 at 10 33 09 AM

Right... `es-site` needs the css 🙃 . I fixed the issue where seeing and made the changes "less breaking". I will name Mike as a reviewer as well.

mpleroux (approved):
I reviewed the changes and they look fine to me. I also ripped the nav and checked it in es-site. I didn't spot any regressions.

You had initially removed this old selector before putting it back. I originally added it to fix some problem and I don't think it's still necessary:

```scss
// Override BS4 defaults for consistency
.nav-es-container {
...
    .menu {
        ....
        ul, li {
            display: block;
            width: 100%;
            border-style: none;
        }
    }
}
```

Your addition of `width: 100%` to `.nav-item` fixed the problem Eric mentioned.

I examined the `ul` and `li` elements inside `.dropdown-list` which is used by the mobile dropdown menus. You can get to it by inspecting "Tips & guides" then looking for the element with the class `dropdown-list`.

- In the `ul`, the width is overridden by `w-auto` in the HTML, the border style doesn't clear anything, and changing `display` from `block` back to `flex` doesn't have any noticeable effect
- In the `li` elements, those three styles are overridden by others with higher specificity

So I think you can get rid of that selector if you like.

I checked the ripped nav with WAVE and it reported one minor accessibility issue that's unrelated to your PR. It's caused by using the same ID for mobile and desktop form elements. Those IDs need to be refactored anyway because they're using the text of the menu items, spaces and all.

```txt
Multiple form labels
A form control has more than one label associated with it.
```

PR #963, opened 2023-06-14
feat: Add error page component

suleiy:
### 🔗 Linked issue
https://energysage.atlassian.net/browse/CED-311
### ❓ Type of change

### 📚 Description

- Added `EsErrorPage` component that can be used in nuxt apps error pages
- Create three new SVGs: `IconContactUs`, `IconMarketplace`, `IconInfoBlogPost`

### 🥼 Testing

- Updated test snapshots in es-vue-base/tests

### 📝 Checklist


es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,107 @@
+<template>
+    <div class="my-5 container">
+        <div class="mb-5 mx-auto col-md-7 text-center">
+            <slot name="errorType" />
```
lgeggleston, line 4: While using slots is definitely the way to go for most design system components that have text/content like this, I think for the purpose of this it might make sense to use prop(s) instead.  The reason would be to avoid having to repeat this text on each repo this is dropped into, since the text and error types aren't likely to change.

To that end, I think probably only the error code itself would need to be a prop - The message texts themselves could just be data in this component, mapped to the error codes;  and if the input is an error code other than these 4, could map to a default generic msg.

es-design-system/pages/molecules/es-error-page.vue
```diff
@@ -0,0 +1,105 @@
...
+<template>
+    <div>
+        <h1>
+            EsErrorPage component
+        </h1>
+        <p>
+            The Error Page component is intended to be used in error pages across nuxt apps.
+        </p>
+        <div class="my-450">
+            <h2>
```
lgeggleston, line 11: Probably one example is enough for the documentation page - when people look at the template for this page under "Documentation" they'll be able to see how it's customized with i.e. a prop.  You could add a line in the description up top though about the 4 error codes explicitly handled?

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,107 @@
...
+            <slot name="errorMsg" />
+        </div>
+
+        <div class="row">
+            <a
+                :class="linkClass"
+                :style="linkStyle"
+                href="mailto:hello@energysage.com">
+                <div class="circle mb-2">
```
lgeggleston, line 13: Small thing: The color of the circles in figma are the default background color for nuxt pages, so I'm not if there are supposed to be visible circles or not.  It looks fine as-is, but could be worth a quick check with Maria DiPietro if those circles are meant to be visible?

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,107 @@
+<template>
+    <div class="my-5 container">
```
lgeggleston, line 2: Ah this is one change from the version of this page on `es-site` - we use bootstrap-vue, which means instead of `class="container"` the div will be a `<b-container>` component.  Also will replace BContainer, BRow, and BCol, [documentation here](https://bootstrap-vue.org/docs/components/layout)

lgeggleston, line 2: For each of the spacing classes (things like `my`, `mb`, `pb`), let's also change these to updated design system classes - should be a direct mapping from the "Deprecated Spacing" section [here](https://design.energysage.dev/atoms/spacing)

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,107 @@
...
+    name: 'EsErrorPage',
+    components: {
+        IconCommunitySolar,
+        IconContactUs,
+        IconMarketplace,
+        IconInfoBlogPost,
+    },
+    data() {
+        return {
+            linkClass: 'col-6 col-md-3 text-center',
```
lgeggleston, line 87: Generally, I think data() is probably not the place for repeated style classes like this.  It's a great instinct to have as little repeated code as possible, but it's also okay to just repeat bootstrap style classes like in the es-site page.  Or alternatively, computed properties might be a better place to reuse classes than data.

If you want though, you could also see if it works to refactor these links as a `v-for`, and use data to instead keep a list of links+icons to loop through!  It's not necessary but might also be a satisfying option.

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,107 @@
...
+            linkStyle: 'display:flex; align-items: center; flex-direction:column',
+            iconFill: '100%',
+        };
+    },
+};
+</script>
+<style lang="scss" scoped>
+@import '~@energysage/es-bs-base/scss/includes';
+
+.circle {
```
lgeggleston, line 98: in the same vein of repeated styles, most of these can be bootstrap utility classes (i.e. `align-items-center`; `bg-gray-100`; `border-circle`).  Leaning towards putting those as classes and leaving css as minimal as possible

lgeggleston (commented):
Hi Su Lei, this is some really good work!  Thanks for jumping into a new repo and dev process so quickly.  My biggest suggestion is probably refactoring this to pass in the error code as a prop rather than using slots, and then there's a few smaller ones.  Let me know if you have questions!

suleiy:
Hello Lia, thank you for the detailed feedback! I have made changes to pass in the error code as a prop instead of using slots, and other suggested changes (using boostrap-vue, spacing classes, bootstrap utility classes, etc). I also talked to Maria and she suggested to use white circles instead of the of gray and I have made changes to accommodate that as well. Please let me know if there are suggestions regarding to the changes made. Thanks again!

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,93 @@
+<template>
+    <b-container class="my-450">
+        <b-col
```
lgeggleston, line 3: This was on the original bootstrap page too and doesn't seem to be a problem , but a `<b-row>` around this `b-col` would be good practice.

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,93 @@
...
+        return {
+            errorCode: {
+                // eslint-disable-next-line max-len
+                err403: "Oops! If you're seeing this message, there's a good chance you have cookies or referrers turned off in your browser.",
+                err404: 'Oops! We are terribly sorry, but there is nothing bright to see here.',
+                err500: 'Oops! We are terribly sorry, but our server is not so bright today. Please try again.',
+                err503: 'Service unavailable: The page you are requesting is temporarily unavailable.',
+            },
+            genericMsg: 'Oops! Something went wrong!',
+            iconList: ['IconContactUs', 'IconCommunitySolar', 'IconInfoBlogPost', 'IconMarketplace'],
```
lgeggleston, line 78: These 3 lists could probably be consolidated into one list of objects such as `[{icon: , link: , label: }, ... ]`

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,93 @@
+<template>
+    <b-container class="my-450">
+        <b-col
+            md="7"
+            class="my-450 mx-auto text-center">
+            <h1 class="font-weight-boldest display-3">
+                {{ errorType }}
+            </h1>
+            <p
```
lgeggleston, line 9: The message should still be a `<h2>` rather than a `<p>`

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,93 @@
...
+                err503: 'Service unavailable: The page you are requesting is temporarily unavailable.',
+            },
+            genericMsg: 'Oops! Something went wrong!',
+            iconList: ['IconContactUs', 'IconCommunitySolar', 'IconInfoBlogPost', 'IconMarketplace'],
+            linkList: ['mailto:hello@energysage.com', 'https://www.energysage.com/solar/', 'https://news.energysage.com/', 'https://www.energysage.com/market-intro/'],
+            labelList: ['Contact Us', 'Learn About Solar', 'Informative Blog Post', 'EnergySage Marketplace'],
+        };
+    },
+    computed: {
+        error() {
```
lgeggleston, line 84: This works fine, but not sure it's necessary since the prop is already a string - you could also use the prop like `'404'` directly as keys for `errorCode`

lgeggleston (commented):
Great job making these changes, especially removing css and refactoring into a `v-for` for the icon columns!!  There are just a couple of tweaks left (first 2 comments); p -> h2 the only necessary one.  
When you're ready, go ahead and tag Ryan & Tom Leo for a final check on this. Thanks!

suleiy:
Thank you Lia! I have made changes to switching the error message from p -> h2 and also other suggested changes. 
Hello @RyanMulready and @tomleo, could you please provide some final check/feedback on this ticket? Thank you!

lgeggleston (approved):

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -0,0 +1,113 @@
...
+                    link: 'https://news.energysage.com/',
+                    label: 'Informative Blog Post',
+                },
+                {
+                    icon: 'IconMarketplace',
+                    link: 'https://www.energysage.com/market-intro/',
+                    label: 'EnergySage Marketplace',
+                },
+
+            ],
```
RyanMulready, line 104: For these link values can you follow a [similar](https://github.com/EnergySage/es-ds/blob/6ee826c3f953034bbb26f36a48952cac5faa5039/es-vue-base/src/lib-utils/footer-content.js#L6) pattern to how our [footer](https://github.com/EnergySage/es-ds/blob/c26ec0da52be9c05fafafe2213564f55d7768c86/es-vue-base/src/lib-components/EsFooter.vue#L126) is doing it?

We want to be able to control the links per environment and make maintaining them easier.

es-vue-base/src/lib-icons/icon-contact-us.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+    <svg
+        :style="{
+            height: height,
+            width: width,
+        }"
+        viewBox="0 0 48 31"
```
RyanMulready, line 7: The `viewBox` of our icons should be relatively consistent. Most are expected to be `0 0 24 24`, with some exceptions being `0 0 32 32`.

Unfortunately the icons here are nether of those and are not uniform amongst themselves. Please ask design for corrected sizes in one of those two viewBox definitions.

es-design-system/components/ds-molecules-list.vue
RyanMulready, line 1: This should be an organism not a [molecule](https://atomicdesign.bradfrost.com/chapter-2/). Can you please move `EsFooter` as well?

RyanMulready (changes requested):
Display looks good, nice job 🚀 . Some implementation changes that need to be made inline.

suleiy:
Hello Ryan, thank you for the feedback! I have made changes to make EsErrorPage and EsFooterPage as organisms. The icons also now have viewbox `0 0 32 32`. 
I also created a new [`error-page-content.js`](https://github.com/EnergySage/es-ds/blob/CED-311-error-page-main-contents/es-vue-base/src/lib-utils/error-page-content.js), and export it as [`getEsErrorPageContent`](https://github.com/EnergySage/es-ds/blob/dc29840856a62caa03c4c50c8914c3a5c30a4935/es-vue-base/src/lib-utils/index.js#L2C21-L2C42) in index.js and import it in [`EsErrorPage`](https://github.com/EnergySage/es-ds/blob/dc29840856a62caa03c4c50c8914c3a5c30a4935/es-vue-base/src/lib-components/EsErrorPage.vue#L83C1-L84C1). Would this be how the links should be handled?

RyanMulready (approved):
Changes look great now, nice job 🔥 . In the future when you are ready for a "re-review" click the circle icon next to the persons name. This will alert them with a notification and you're likely to get it looked at sooner than just a comment.

image

PR #967, opened 2023-06-15
CORE-4327-add-slot-es-form

srazh:
### 🔗 [Linked issue](https://energysage.atlassian.net/browse/CORE-4552)

CORE-4327-add-slot-es-form
### ❓After initial review, the premise of the ticket has changed. See resolved comments.

- [x ] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description
- refactor logic so we only use a slot to display content
- added a primary es-form-msg example that displays content 
- ensured all form messages hide after 20 seconds

### 🥼 Testing

- Go to localhost:8500 and navigate to Molecules, then EsFormMsg. View the newly added form message type by pressing "show default message"

### 📝 Checklist

- [ x] I have linked an issue or discussion.
- [x ] I have updated the documentation accordingly.
- [ x] I have documented testing approach


es-vue-base/src/lib-components/EsFormMsg.vue
```diff
@@ -1,19 +1,24 @@
...
         <div class="d-flex">
             <div class="pr-50">
                 <IconCircleAlert v-if="variant === 'danger'" />
                 <IconCircleCheck v-if="variant === 'success'" />
             </div>
-            {{ message }}
+            <template v-if="!hasSlot">
+                {{ message }}
+            </template>
+            <template name="slotContent">
```
jimlind, line 19: What is the warning about?

jimlind:
You branch is out of date. Can you merge in main?

es-vue-base/src/lib-components/EsFormMsg.vue
srazh: The "warning" is a link.

ericdouglaspratt (commented):
Was doing some ESDS reviews and had a suggestion!

ericdouglaspratt:
@srazh Can you commit the updated snapshot file? Happy to give this a review after that.

es-vue-base/src/lib-components/EsFormMsg.vue
```diff
@@ -46,34 +50,42 @@ export default {
...
             default: 'danger',
             required: false,
-            validator: (val) => ['danger', 'success'].includes(val),
+            validator: (val) => ['danger', 'success', 'primary'].includes(val),
         },
     },
     data() {
         return {
             dismissCountDown: this.timeout,
+            messageType: '',
```
ericdouglaspratt, line 59: It looks like this isn't used - might be okay to remove this?

es-vue-base/src/lib-components/EsFormMsg.vue
```diff
@@ -1,19 +1,21 @@
...
         <div class="d-flex">
             <div class="pr-50">
                 <IconCircleAlert v-if="variant === 'danger'" />
                 <IconCircleCheck v-if="variant === 'success'" />
+                <IconInfo v-if="variant === 'primary'" />
             </div>
-            {{ message }}
+            <slot>
+                {{ formMsg }}
+            </slot>
```
ericdouglaspratt, line 18: This should be able to just be `<slot />` with nothing inside. And any message can just be passed into the slot when this component is used.

es-vue-base/src/lib-components/EsFormMsg.vue
```diff
@@ -23,16 +25,18 @@ import {
...
     name: 'EsFormMsg',
     components: {
         BAlert,
         IconCircleAlert,
         IconCircleCheck,
+        IconInfo,
     },
     props: {
-        message: {
+        formMsg: {
```
ericdouglaspratt, line 39: Can we remove this and just rely on the slot to pass in the message?

ericdouglaspratt (changes requested):
Summarized some initial thoughts we discussed on Slack in here.

es-design-system/pages/molecules/es-form-msg.vue
```diff
@@ -8,17 +8,35 @@
...
-        <es-button @click="fakeFormMsg(false)">
-            Show Error
+        <es-button
+            @click="showInfo = true">Show Info
         </es-button>
+
+        <es-form-msg
+            class="my-450"
+            :show="showSuccess"
+            :variant="'success'"
```
ericdouglaspratt, line 25: We can simplify this to `variant="success"` since it's just a string.

es-design-system/pages/molecules/es-form-msg.vue
```diff
@@ -8,17 +8,35 @@
                 bootstrap-vue alert
             </b-link>
         </p>
-        <es-button @click="fakeFormMsg()">
-            Show Success
+
+        <es-button
+            @click="showSuccess = true">Show Success
```
ericdouglaspratt, line 13: Nit: Can we change this to "Show success" to match our [writing guidelines about EsButton captions](https://design.energysage.dev/molecules/es-button) being sentence case?

es-design-system/pages/molecules/es-form-msg.vue
```diff
@@ -8,17 +8,35 @@
...
-            @hidden="formMsg = ''" />
+            :show="showError"
+            :variant="'danger'"
+            @hidden="showError = false">The server responded with an error and we were unable to complete your request. Please try again
+        </es-form-msg>
+        <es-form-msg
+            class="my-450"
+            :show="showInfo"
+            :variant="'primary'"
+            @hidden="showInfo = false">Need Expert Advice? &nbsp;<a href="https://www.energysage.com/"> Schedule a free call with our Energy Advisor Team!</a>
```
ericdouglaspratt, line 38: Nit: can we change this to "Need expert advice?" so messaging copy is consistently in sentence case?

"Energy Advisor" is a brand term so that should still be capitalized, but "team" should be lowercase, as well.

es-vue-base/src/lib-components/EsFormMsg.vue
```diff
@@ -1,19 +1,18 @@
 <template>
     <b-alert
-        v-if="message"
-        :show="dismissCountDown"
-        :variant="variant"
-        dismissible
-        class="form-msg"
+        :show="dismissCountDown "
```
ericdouglaspratt, line 3: Extra space at the end here that can be removed

ericdouglaspratt (changes requested):
Two things that broke due to these changes that need to be fixed:
1. `/examples/form-validation` uses `EsFormMsg` so we just need to update that to pass in its message via the slot and use booleans, just like we did with the `es-form-msg.vue` docs page.
2. The Jest unit test for EsFormMsg at `/tests/EsFormMsg.spec.js`, as we can see in the snapshot, no longer tests anything correctly. It also needs to be updated to use the slot and booleans and should successfully render the message it's trying to pass in.

es-vue-base/src/lib-components/EsFormMsg.vue
```diff
@@ -23,57 +22,69 @@ import {
...
+
+        show: {
+            type: Boolean,
+            default: false,
         },
+
         timeout: {
             type: Number,
-            default: 20,
+            default: 5,
```
ericdouglaspratt, line 44: One thing I realized - we should probably keep this default at 20 seconds for the EsFormMsg component itself, otherwise forms in the real world that use this might have their errors disappear before customers have a chance to fully read them.

es-vue-base/tests/EsFormMsg.spec.js
```diff
@@ -8,7 +8,7 @@ describe('EsFormMsg', () => {
         const wrapper = mount(EsFormMsg, {
             ...jestVue,
             propsData: {
-                message: 'Test form msg',
+                show: true,
```
ericdouglaspratt, line 11: This is fine as a basic test of the EsFormMsg rendering, but it doesn't quite keep the same functionality as before where `'Test form msg'` was rendered as the message inside it, which is a test more in line with how it'll be used. `EsButton.spec.js` has a good example where you can pass in a `slots` object just before `propsData` here to pass in the message.

ericdouglaspratt (changes requested):
The behavior of `form-validation.vue` isn't quite right - I'm seeing the error message show immediately while the button is still spinning, while the [original version](https://design.energysage.dev/examples/form-validation) has the button spin for a while and then shows the error message upon completion of the (fake) form submission.

If you look at the `fakeServerRequest` method, it's using the `showFormError()` function from `lib-mixins/forms.js`, which sets `formMsgVariant` and `formShowError`. We should be using those variables for `variant` and `show` rather than hardcoded `'danger'` and `showError`, respectively - then it should work.

es-design-system/pages/molecules/es-form-msg.vue
```diff
@@ -8,17 +8,42 @@
...
-            Show Error
+        <es-button
+            @click="showInfo = true">
+            Show info
         </es-button>
+
+        <es-form-msg
+            class="my-450"
+            :show="showSuccess"
+            :variant="success"
```
ericdouglaspratt, line 28: I'm getting console warnings and the EsFormMsg docs page examples are the wrong color. We'll need to change this to `variant="success"` because with the colon in front, it's looking for a variable named `success` and not finding it, so it's passing in `undefined` and EsFormMsg is defaulting to `danger` because that's the default value of the `variant` prop.

es-design-system/pages/molecules/es-form-msg.vue
```diff
@@ -8,17 +8,42 @@
...
+            @hidden="showSuccess = false">
+            Saved successfully
+        </es-form-msg>
         <es-form-msg
             class="my-450"
-            :variant="formMsgVariant"
-            :message="formMsg"
-            @hidden="formMsg = ''" />
+            :show="showError"
+            :variant="danger"
```
ericdouglaspratt, line 35: Same thing here - we need to remove the colon in front so this is passed in as a string rather than a nonexistent variable.

es-design-system/pages/molecules/es-form-msg.vue
```diff
@@ -8,17 +8,42 @@
...
+            @hidden="showError = false">
+            The server responded with an error
+            and we were unable to complete your request. Please try again
+        </es-form-msg>
+        <es-form-msg
+            class="my-450"
+            :show="showInfo"
+            :variant="primary"
```
ericdouglaspratt, line 43: Same thing here - we need to remove the colon in front so this is passed in as a string rather than a nonexistent variable.

es-vue-base/src/lib-mixins/forms.js
```diff
@@ -5,8 +5,9 @@ export default {
     data() {
         return {
             submitInProgress: false,
-            formMsg: '',
-            formMsgVariant: 'danger',
+            formShowSuccess: false,
+            formShowError: false,
+            formMsgVariant: '',
```
ericdouglaspratt, line 10: We should keep `'danger'` as the initial value for `formMsgVariant`, otherwise an empty string could be passed in, which would fail the prop validation set up [here](https://github.com/EnergySage/es-ds/pull/967/files#diff-c8e866b38f24e2b2f5834114c4ff3f537ab64df09137064efa1add7e42ec0025R49) and generate a warning in the console. The default value would still be applied but worth avoiding potentially confusing console warnings for people trying to use the default forms mixin behavior.

es-design-system/pages/examples/form-validation.vue
```diff
@@ -37,10 +37,13 @@
                 <b-form
                     @submit.stop.prevent="onSubmit">
                     <es-form-msg
-                        class="mt-100"
+                        class="my-450"
```
ericdouglaspratt, line 40: We can probably keep the `mt-100` rather than the `my-450` here. The latter gives a ton of unnecessary space around the error message:
Screen Shot 2023-07-13 at 2 35 10 PM

ericdouglaspratt (changes requested):
Jest test looks good! There are a couple functionality things that aren't working though - and some console warnings we should fix.

PR #1058, opened 2023-07-17
feat: add EsVideo component

suleiy:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-643
### ❓ Type of change

### 📚 Description

- Create a `EsVideo` component extracted from [VideoWithCover component ](https://github.com/EnergySage/es-cms-pages/blob/main/components/VideoWithCover.vue )in`es-cms-pages`
- Add `IconVideoPlay` under base icons
Note: added !important to text-colors mixin in [scss/utilities/_text.scss](https://github.com/EnergySage/es-ds/pull/1058/files#diff-b52011214dae30d05a5e349032b7574eef37585a7bc9d2cc25e7cec4116bd403), and might have to check to make sure text colors aren't changing unexpectedly.

### 🥼 Testing

Unit tests under es-vue-base/tests/

### 📝 Checklist


es-design-system/pages/molecules/es-video.vue
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+        <ds-doc-source
+            :comp-code="compCode"
+            comp-source="es-vue-base/src/lib-components/EsVideo.vue"
+            :doc-code="docCode"
+            doc-source="es-design-system/pages/molecules/es-video.vue" />
+    </div>
+</template>
+<script>
+import placeholderEsThumbnail from '@/assets/placeholder/energysage_thumbnail.png';
```
suleiy, line 25: I have the video thumbnail icon currently stored under the assets/placeholder folder, but I am not sure where it should be stored?

es-vue-base/src/lib-icons/icon-video-play.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+    <svg
+        :style="{
+            height: height,
+            width: width,
+        }"
+        viewBox="0 0 74 54"
```
suleiy, line 7: Would the viewBox values need to be changed to be consistent with other base icons or would this icon be in a different category? The video play icon for the EsVideo in es-cms-pages is 74x54px.

es-vue-base/src/lib-components/EsVideo.vue
```diff
@@ -0,0 +1,91 @@
...
+            allowfullscreen
+            aspect="16by9"
+            :src="autoplayUrl"
+            :title="altText"
+            type="iframe" />
+        <es-button
+            v-else
+            class="VideoWithCover-button border-0 p-0 position-relative text-gray w-100"
+            @click="showVideo = true">
+            <b-img
```
lgeggleston, line 14: Since it would be ideal to still allow folks to use `<storyblok-image>` in the Nuxt repos for this, what do you think about using `<slot>` for the image here, with this `b-img` inside it as a fallback?
- see `EsSupport` for the example of this pattern
- probs remove the `sizes` attribute on the b-img, I think that's a nuxt-img prop
- the image url & alt props would become optional, and the fallback b-img would show only if those props are given & the slot is not

lgeggleston (commented):
Nice work! I wrote one idea to still allow `StoryblokImage` using a slot the way EsSupport does, but other than that the docs page & everything look good to me.

suleiy:
Thank you Lia! I've made changes to include slots to allow using`StoryblokImage`. 
For the .text-gray utility class, I added !important to text-colors mixin in [scss/utilities/_text.scss](https://github.com/EnergySage/es-ds/pull/1058/files#diff-b52011214dae30d05a5e349032b7574eef37585a7bc9d2cc25e7cec4116bd403), and it works for this ticket when I run locally. I looked at some other components which uses text-colors class and it did not cause any style errors for them, but I wanted to make sure if this change might effect others that might not want the !important?

es-design-system/pages/molecules/es-video.vue
```diff
@@ -0,0 +1,51 @@
...
+
+export default {
+    name: 'EsVideoDocs',
+    data() {
+        return {
+            compCode: '',
+            docCode: '',
+            altText: 'Community solar: Everything you want to know before you join',
+            placeholderEsThumbnail,
+            embedUrl: 'https://www.youtube.com/embed/hgmZG3GLLNg',
```
ericdouglaspratt, line 35: Since `altText` and `embedUrl` are only used once, I'd say it might be cleaner to move these values straight onto the component definition above, and then we won't need these variables in `data()`.

es-vue-base/src/lib-components/EsVideo.vue
```diff
@@ -0,0 +1,95 @@
...
+    components: {
+        BImg,
+        BEmbed,
+        IconVideoPlay,
+    },
+    props: {
+        // eslint-disable-next-line vue/require-default-prop
+        altText: {
+            type: String,
+            required: false,
```
ericdouglaspratt, line 47: I think for page load performance purposes, we're going to want a cover image in all cases, so we should probably make `altText` and `coverImageUrl` required.

es-design-system/pages/molecules/es-video.vue
```diff
@@ -0,0 +1,51 @@
+<!-- eslint-disable vuejs-accessibility/no-static-element-interactions -->
+<template>
+    <div>
+        <h1>
+            Video
+        </h1>
+        <p>
+            Shows the embed video when the cover is clicked.
```
ericdouglaspratt, line 8: Since our URL parsing logic assumes the provided embed URL is a YouTube embed URL, I'd recommend making this a little more specific, like "Shows a YouTube video when the provided cover image is clicked."

ericdouglaspratt (changes requested):
Nice work! A couple of comments.

suleiy:
Thank you @ericdouglaspratt! I've made changes to the icon, cover image url, and other changes to the EsVideo.vue and es-video.vue files.

es-design-system/pages/atoms/icons.vue
```diff
@@ -481,6 +477,20 @@
                 <code>IconFilePdf</code>
             </li>
         </ul>
+        <h2>
+            Miscellaneous Icons
+        </h2>
+        <p>
+            Default size for file icons is 74px by 54px.
```
ericdouglaspratt, line 484: The icons in this section will probably not end up being consistent sizes - we can probably remove this note so we don't need to figure out how to update it when we add a new one.

es-design-system/pages/molecules/es-video.vue
```diff
@@ -1,18 +1,21 @@
...
-            <es-video
-                :alt-text="altText"
-                class="my-450 mb-200 mb-lg-0"
-                :cover-image-url="placeholderEsThumbnail"
-                :embed-url="embedUrl" />
+            <b-row class="justify-content-center">
+                <b-col md="8">
+                    <es-video
+                        alt-text="Community solar: Everything you want to know before you join"
+                        class="my-450 mb-200 mb-lg-0"
```
ericdouglaspratt, line 14: Couple small things:
- The spacing looks great on desktop but is a bit large on mobile - maybe we can do `my-200 my-lg-450`?
- Because of the `my-450`, I don't think the `mb-200 mb-lg-0` is doing anything, since `my` covers both `mb` and `mt`, so we can probably remove those.
- To make the `<es-video>` example easier to copy out for any use case, might be worth moving the margin class(es) up to the `<b-row>` element.

ericdouglaspratt (approved):
Looks great! A few small suggestions but otherwise okay to merge.

suleiy:
Fixed the icon description and the margins, thank you!

ericdouglaspratt (approved):

PR #1089, opened 2023-08-02
CED-800: update EsAccordion to use heading tags

suleiy:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-800

### ❓ Type of change

### 📚 Description

- Update EsAccordion to use heading tags
    - Accepts a headingTag parameter that defaults to 'h3'
    - Includes a <button> tag,  positioned over on top of the 'h3' tag  and contains hidden text to show 'collapse' or 'expand'

### 🥼 Testing

- Updated the EsAccordionList.spec.js test file

- I'm unsure about how the button should include hidden text. Currently, I have the 'collapse/expand' texts in a span and hidden display.

### 📝 Checklist


es-vue-base/src/lib-components/EsAccordion.vue
```diff
@@ -1,18 +1,24 @@
...
+                    'EsAccordion-heading--visible': isVisible,
+                }">
                 <slot name="title" />
                 <icon-chevron-down class="EsAccordion-icon flex-shrink-0 ml-200" />
-            </es-button>
+            </component>
+            <button
+                class="EsAccordion-button position-absolute w-100 h-100 border-0 bg-transparent"
+                @click="handleClick">
+                <span class="d-none">{{ isVisible ? 'collapse' : 'expand' }}</span>
```
ericdouglaspratt, line 20: If we replace the `d-none` here with the `sr-only` class, this text will be readable by screen readers, but still visually hidden, via some `position: absolute` magic. Screen readers don't read out elements that are `display: none`.

es-vue-base/src/lib-components/EsAccordion.vue
```diff
@@ -1,18 +1,24 @@
...
     <div class="EsAccordion border-bottom border-light rounded-bottom">
-        <header role="tab">
-            <es-button
-                class="EsAccordion-button h-auto align-items-center d-flex font-weight-bold justify-content-between px-100 px-sm-200 py-100 rounded-0 text-body text-left text-decoration-none w-100"
+        <header
+            role="tab"
+            class="position-relative d-inline-block w-100">
+            <component
+                :is="headingTag"
+                class="EsAccordion-heading mb-0 h-auto align-items-center d-flex font-weight-bold justify-content-between px-100 px-sm-200 py-100 rounded-0 text-body text-left text-decoration-none"
```
ericdouglaspratt, line 8: A couple things here:
- Some of these utility classes were only here to override certain EsButton styles (at least `h-auto`, `font-weight-bold`, `rounded-0`, `text-left`, `text-decoration-none`, I believe), and we may no longer need them. Can we remove the unnecessary ones?
- Changing the default to an `<h3>` is appropriate, but it does dramatically change the rendered text size on desktop. [CED-799](https://energysage.atlassian.net/browse/CED-799) will solve for the ability to pass in a `variant` prop to customize the font size, among other styling, but in order to minimize breaking visual changes in this PR (and set up what will be needed for the `'default'` variant, let's set `font-size-100` here to keep it the same size it was before.

es-vue-base/src/lib-components/EsAccordion.vue
```diff
@@ -1,18 +1,24 @@
 <template>
     <div class="EsAccordion border-bottom border-light rounded-bottom">
-        <header role="tab">
-            <es-button
-                class="EsAccordion-button h-auto align-items-center d-flex font-weight-bold justify-content-between px-100 px-sm-200 py-100 rounded-0 text-body text-left text-decoration-none w-100"
+        <header
+            role="tab"
+            class="position-relative d-inline-block w-100">
```
ericdouglaspratt, line 5: Would removing the `d-inline-block w-100` still work, as I believe the `<header>` tag would render at its default `display: block` and go the full width? If so, could be simpler.

es-design-system/pages/molecules/es-accordion.vue
```diff
@@ -353,11 +349,15 @@ export default {
                         EsAccordionList to signify that this item should start out expanded.
                     `,
                 },
+                {
+                    name: 'headingTag',
+                    description: `
+                        Defaults to h3. The tag used can be customized as needed.
+                        For example, if the accordion is used within a section that is already under an <h3>, it may need to be an h4.
```
ericdouglaspratt, line 356: I'm seeing a line length warning on this line, probably just need to line break mid-way through.

ericdouglaspratt (changes requested):
Nice work! A couple things.

suleiy:
Made changes to add transition to the header and also add/remove classes, thanks!

ericdouglaspratt (changes requested):
Looks awesome - just one item on reduced motion and this should be good to go.

ericdouglaspratt (approved):
Nice work!

PR #1131, opened 2023-08-08
refactor: allow constrained es-support-card

lgeggleston:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-635 (Change will then be used in `es-cms-articles`)

### ❓ Type of change

### 🥼 Testing

- Tested appearance in all viewports; current default examples are unaffected and constrained example looks reasonable.

- The Figma design for the constrained version of this module does look slightly different; it uses a 75x75px version of the image which is also not centered.  However due to the way this image display is implemented, the fact we allow a slot for it, and the mobile version showing the 100px image regardless, I thought that particular difference would require more refactoring than it's worth.  If that's not the case, I can log a ticket to investigate matching more exactly.

### 📝 Checklist

### DIFF
--- a/es-design-system/pages/organisms/es-support-card.vue
+++ b/es-design-system/pages/organisms/es-support-card.vue
@@ -59,9 +59,74 @@
         </div>
         <div class="mb-450">
             <h2>
+                Constrained
+            </h2>
+            <p class="mb-200">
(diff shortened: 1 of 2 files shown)


mpleroux (approved):
This looks like a good compromise. I guess we could center the constrained version in very wide viewports?

PR #1136, opened 2023-08-09
feat: zip code form hide privacy policy prop

lgeggleston:
### 🔗 Linked issue

To be used in https://energysage.atlassian.net/browse/CED-815

### ❓ Type of change

### 📚 Description

- Add a `hidePrivacyPolicy` prop that defaults to false on `EsZipCodeForm`.  Having the privacy policy is important for EEAT purposes and this is only for a specific design use case (Sticky zip with limited height on solar companies), so added appropriate warnings in comment and docs page. 
- The [design](https://www.figma.com/file/zRKAY3boW1vcLCRaUZPgyU/Solar-Companies-Project-Refresh?type=design&node-id=2110-589&mode=design&t=mGjnokPHvRAAAhhV-0) this is intended to support:
Screen Shot 2023-08-09 at 10 59 43 AM

### 🥼 Testing

On all viewports; existing examples are unaffected by this change.

-

### 📝 Checklist


ericdouglaspratt:
@lgeggleston I'm wondering something more basic here - with Solar Companies, we're not redesigning the page (yet), just porting it into Storyblok basically, right? If so, would it be simpler to say that rather than modifying the current design system to match the old design from 2022, maybe we could just use the CTA banner from the current design system for that page?

lgeggleston:
> @lgeggleston I'm wondering something more basic here - with Solar Companies, we're not redesigning the page (yet), just porting it into Storyblok, right? If so, would it be simpler to say that rather than modifying the current design system to match the old design from 2022, maybe we could just use the CTA banner from the current design system for that page?

@ericdouglaspratt Yep that's right, we're just putting the old design in Storyblok but subbing in design system elements as possible.  My main concern about using the current banner or zipCodeForm was this CTA is sticky, so it looks strange if too thick and those options would add height.  For context, I talked with Huanran yesterday about the options for this page and he landed on that small redesign; but I'll check about the EsCtaBanner alternative

PR #1139, opened 2023-08-11
fix: Use Fetch instead of Axios

zacharyjarnagin:
### 🔗 Linked issue

#113

### ❓ Type of change

### 📚 Description

Originally, the es-file-input component was using the axios instance defined in nuxt. This caused issues when the component was used in CSM since the axios instance in nuxt had a token on every POST request, causing S3 uploads to fail.

Ideally, we could have just created a new nuxt instance, but axios is not a dependency in es-design-system. To get around this, the component is now using the fetch module. The downside is obtaining upload progress is extremely convoluted, so it has been removed.

Removing the uploadProgress event meant I had to modify EsFileThumbnail so the user can see a normal spinner whenever the percent loaded is zero. The percent loaded for a file will now only be either zero or 100, but the functionality for upload progress can easily be added back in.

### 🥼 Testing

- Manual testing of uploading to S3. I had overmind running and posted to olympus by going to `localhost:8104/api` and creating a file, getting the fileUploadUrl, and adding the fields manually under the `onSubmit` method with the key `additionalFields`. Once I did that, I submitted the file and downloaded it from Olympus

### 📝 Checklist


RyanMulready (approved):
Code changes look good to me. Can you log a ticket in `ES-DS` around adding back tracking progress? You can leave it unassigned

zacharyjarnagin:
> Code changes look good to me. Can you log a ticket in `ES-DS` around adding back tracking progress? You can leave it unassigned

Added as #161!

es-vue-base/src/lib-components/EsFileThumbnail.vue
gigiwargin, line 1: My only nit would be that the docs for this loading state are now a little funky. By default, the spinner is going. When "see loading state" is hit, it shows the progress circle. Might want to do two?

gigiwargin (approved):
Works like a charm! Nice job Zach 🥳  One comment inline but doesn't block approval!

es-vue-base/src/lib-components/EsFileThumbnail.vue
gigiwargin: I agree that it's the intended behavior! But if I didn't know there were two loading states I might be confused looking at the docs 🤷‍♀️

zacharyjarnagin: I changed the button from "Start loading state" to "Start file upload" to try and highlight that the percentage is shown when a file is in the middle of uploading

PR #1146, opened 2023-08-15
CSM-2802: Add Spacing Between Review Star Rating

satwikgoyal:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CSM-2802

### ❓ Type of change

### 📚 Description

- Before the change was made, the review column and ratings column in ReviewModal had no space in between. After the changes in this PR, there is a 0.75rem right padding in the ratings column in ReviewModal. 

### 🥼 Testing

- Check EsReviewList in Organisms. Click on 'View All Ratings' and inspect the modal. There will be a padding-right (=0.75rem) in the ratings column

- Does the spacing seems to be right (should it be less or more)?

### 📝 Checklist

### DIFF
--- a/es-vue-base/src/lib-components/EsReviewModal.vue
+++ b/es-vue-base/src/lib-components/EsReviewModal.vue
@@ -14,7 +14,7 @@
             class="modal-container">
             <b-row>
                 <b-col
-                    class="p-0"
+                    class="p-0 pr-3"
                     cols="12"
                     md="8"
                     order="2"
--- a/es-vue-base/src/lib-components/EsReviewModal.vue
+++ b/es-vue-base/src/lib-components/EsReviewModal.vue
@@ -14,7 +14,7 @@
             class="modal-container">
             <b-row>
                 <b-col
-                    class="p-0 pr-3"
+                    class="p-0 pr-lg-200"
                     cols="12"
                     md="8"
                     order="2"


es-vue-base/src/lib-components/EsReviewModal.vue
```diff
@@ -14,7 +14,7 @@
             class="modal-container">
             <b-row>
                 <b-col
-                    class="p-0"
+                    class="p-0 pr-3"
```
ericdouglaspratt, line 17: Two things here:
1. The single-number naming scheme for spacing classes is [deprecated](https://design.energysage.dev/atoms/spacing); you'll want to use `pr-100` here instead.
2. You probably don't need the extra padding on mobile; only at the `lg` breakpoint when this section, already starting to appear side-by-side at the `md` breakpoint, is reordered to the left side without any spacing between it and the column to its right. So you can make this `pr-lg-100` so it only starts happening at the `lg` breakpoint and up.

Also, the amount of spacing is up to you, your team, and your designer, but I might recommend doubling it to `pr-lg-200` just to give everything a bit more breathing room.

ericdouglaspratt (changes requested):

es-vue-base/src/lib-components/EsReviewModal.vue
satwikgoyal: @ericdouglaspratt thank you so much for your feedback. I have user `pr-lg-200` as recommended by you. Our designer didn't ask for a particular number, just enough padding to separate the two sub-components so `pr-lg-200` works perfectly!

PR #1170, opened 2023-08-23
feat: toggle programmatic collapse priority for es-collapse

zacharyjarnagin:
### 🔗 Linked issue

[ESDS-162](https://energysage.atlassian.net/browse/ESDS-162)

### ❓ Type of change

### 📚 Description

Adds `prioritizeSuggestedVisible` to the `ESCollapse` component. When false, the `visible` prop is ignored once the user interacts with the collapse. When true, the `visible` prop begins affecting the component even when the user interacts with it.

### 🥼 Testing

- Ensured existing functionality behaves as expected
- Turned on and off the priority toggle, and made sure the behavior was as expected

Nathaniel 👀

### 📝 Checklist

[ESDS-162]: https://energysage.atlassian.net/browse/ESDS-162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/es-design-system/pages/molecules/es-collapse.vue
+++ b/es-design-system/pages/molecules/es-collapse.vue
@@ -8,46 +8,118 @@
                 bootstrap-vue collapse
             </b-link>
         </p>
-        <form>
-            <label for="suggestedVisibleInput">
-                <input
-                    id="suggestedVisibleInput"
-                    v-model="suggestedVisible"
-                    type="checkbox">
-                Toggle collapse programmatically (will be honored until a manual expand or collapse)
-            </label>
-            <br>
-            <label for="prioritizeSuggestedVisibleInput">
-                <input
-                    id="prioritizeSuggestedVisibleInput"
-                    v-model="prioritizeSuggestedVisible"
-                    type="checkbox">
-                Toggle prioritizing programmatic collapse over user interaction
-            </label>
-        </form>
-        <EsCollapse
-            id="testId"
-            :visible="suggestedVisible"
-            :prioritize-suggested-visible="prioritizeSuggestedVisible"
-            class="p-450 my-450"
-            @shown="shownEvent"
-            @toggled="toggledEvent">
-            <template #title>
-                <h2 class="mb-0">
-                    My Title
-                </h2>
-            </template>
-            <!-- eslint-disable max-len -->
+        <div class="mt-450 mb-450">
+            <h2>
+                Default EsCollapse
+            </h2>
             <p>
-                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi in aliquam ex. Nullam vestibulum ex mi, ut suscipit libero condimentum id.
-                Pellentesque eu diam vel nisi molestie porta eget sed odio. Quisque congue risus id metus facilisis, non imperdiet libero rutrum. Mauris
-                vitae ante porttitor, consectetur purus faucibus, euismod ex. Orci varius natoque penatibus et magnis dis parturient montes, nascetur
-                ridiculus mus. Nulla ullamcorper elit sed viverra finibus. Mauris vitae tortor mauris. Cras suscipit nibh nec nisi cursus ornare.
-                Maecenas quis turpis sit amet sapien dapibus sollicitudin viverra eu justo. Vivamus posuere metus sit amet purus tempus volutpat.
-                Donec eleifend elit quam.
+                A normal EsCollapse component. Click it to toggle showing it's contents!
             </p>
-        <!-- eslint-enable max-len -->
-        </EsCollapse>
+            <EsCollapse
+                id="defaultCollapse"
+                class="ml-450 mr-450">
+                <template #title>
+                    <h2 class="mb-0">
+                        My Title
+                    </h2>
+                </template>
+                <!-- eslint-disable max-len -->
+                <p>
+                    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi in aliquam ex. Nullam vestibulum ex mi, ut suscipit libero condimentum id.
+                    Pellentesque eu diam vel nisi molestie porta eget sed odio. Quisque congue risus id metus facilisis, non imperdiet libero rutrum. Mauris
+                    vitae ante porttitor, consectetur purus faucibus, euismod ex. Orci varius natoque penatibus et magnis dis parturient montes, nascetur
+                    ridiculus mus. Nulla ullamcorper elit sed viverra finibus. Mauris vitae tortor mauris. Cras suscipit nibh nec nisi cursus ornare.
+                    Maecenas quis turpis sit amet sapien dapibus sollicitudin viverra eu justo. Vivamus posuere metus sit amet purus tempus volutpat.
+                    Donec eleifend elit quam.
+                </p>
+                <!-- eslint-enable max-len -->
+            </EsCollapse>
+        </div>
+
+        <div class="mt-450 mb-450">
+            <h2>
+                Programmatic EsCollapse
+            </h2>
+            <p>
+                An EsCollapse component that takes a "visible" prop. Click the checkbox to toggle showing it's
+                contents! If you click the collapse itself, the "visible" prop will no longer be honored.
+            </p>
+            <form class="ml-450 mr-450 mt-5">
+                <label for="suggestedVisibleInput">
+                    <input
+                        id="suggestedVisibleInput"
+                        v-model="suggestedVisible"
+                        type="checkbox">
+                    Toggle collapse programmatically (will be honored until a manual expand or collapse)
+                </label>
+            </form>
+            <EsCollapse
+                id="testId"
+                :visible="suggestedVisible"
+                class="p-450"
+                @shown="shownEvent"
+                @toggled="toggledEvent">
+                <template #title>
+                    <h2 class="mb-0">
+                        My Title
+                    </h2>
+                </template>
+                <!-- eslint-disable max-len -->
+                <p>
+                    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi in aliquam ex. Nullam vestibulum ex mi, ut suscipit libero condimentum id.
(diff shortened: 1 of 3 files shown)


tomleo:
@zacharyjarnagin Could you link to the associated JIRA ticket?

zacharyjarnagin:
@tomleo done!

nathanielwarner (changes requested):
I'm unable to get this to work... When I select "Toggle prioritizing programmatic collapse over user interaction", the "Toggle collapse programmatically" checkbox still has no effect after the user has interacted with the collapse.

ericdouglaspratt (changes requested):
I tested it locally and I'm not sure it works for me - I tried toggling the programmatic checkbox a few times first, then manually expanding/collapsing, then checking the new prioritizing checkbox, then toggling the programmatic checkbox and nothing happened.

I'd also suggest a few things:
- Naming the prop in a way that describes the behavior from the user's perspective - maybe something like `disableUserControl` that defaults to `false`?
- Adding the prop to the props table on the docs page
- Separate out the examples for these three different cases (I'm assuming your needs don't include a mid-stream changing of the behavior from user can control it to user can't control it, more that you just need your collapse example to not be able to be interfered with by a user):
     - default example, starts expanded, no programmatic anything, just user can collapse/expand it
     - programmatic example (what it was before this PR), starts expanded, checkbox programmatically toggles, user can interact with it and then the programmatic checkbox stops doing anything
     - new programmatic example, starts expanded, checkbox programmatically toggles, user cannot interact with it

nathanielwarner (approved):
By Zach's instructions, I got it to work by running:
```
rm -rf node_modules && rm -rf */node_modules/ && rm -rf */.nuxt && rm -rf */dist/
make install && make symlink
make dev
```
It may take multiple tries.

nathanielwarner:
It's not a blocking issue (I think), but I noticed that when programmatic collapse is prioritized, the `toggled` event gets emitted, which is supposed to happen when the user toggles it.

zacharyjarnagin:
@nathanielwarner I was able to resolve this by adding an additional data item!

ericdouglaspratt (commented):
Are we worried about this use case?
1. `visible` starts false, collapse starts collapsed (as in our third example on page load)
2. toggle `visible` to true, collapse expands
3. manually collapse the component
4. now we want to programmatically expand it, but `visible` is already true, so we can't do anything except set `visible` to `false` and then set it to `true` again (unless we specifically listen for the `@toggled` event in the parent component and update our `visible` value when the user collapses it)

Starting to feel like these two programmatic examples, taken together, are getting close to the functionality of a simpler `v-model` implementation, especially if we end up needing to listen for the `@toggled` event to update the value of what we're passing in when the user interacts with the collapse, so we can programmatically change it later.

Should we switch to`v-model`, and simply have some kind of `decoupleFromVModelOnUserInput` or `honorUserInteraction` or `isProgrammaticUntilUserInput` boolean prop to accomplish Nathaniel's case of ignoring the programmatic value upon user input?

zacharyjarnagin:
@ericdouglaspratt I believe my latest commit addresses your concern. Let me know if it is what you had in mind!

es-design-system/pages/molecules/es-collapse.vue
```diff
@@ -8,37 +8,118 @@
...
-            @toggled="toggledEvent">
-            <template #title>
-                <h2 class="mb-0">
-                    My Title
-                </h2>
-            </template>
-            <!-- eslint-disable max-len -->
+        <div class="mt-450 mb-450">
+            <h2>
+                Default EsCollapse
```
ericdouglaspratt, line 13: Nit: This could be just "Default", since everything on this page is EsCollapse

es-design-system/pages/molecules/es-collapse.vue
```diff
@@ -8,37 +8,118 @@
...
+            </h2>
+            <p>
+                A normal EsCollapse component. Click it to toggle showing it's contents!
```
ericdouglaspratt, line 16: Typo: "it's" > "its"

es-design-system/pages/molecules/es-collapse.vue
```diff
@@ -8,37 +8,118 @@
...
+                    Maecenas quis turpis sit amet sapien dapibus sollicitudin viverra eu justo. Vivamus posuere metus sit amet purus tempus volutpat.
+                    Donec eleifend elit quam.
+                </p>
+                <!-- eslint-enable max-len -->
+            </EsCollapse>
+        </div>
+
+        <div class="mt-450 mb-450">
+            <h2>
+                Programmatic EsCollapse
```
ericdouglaspratt, line 41: Nit: We could axe the "EsCollapse" here since everything on this page is EsCollapse, and it could be something like "Programmatic, with user override"

es-design-system/pages/molecules/es-collapse.vue
```diff
@@ -8,37 +8,118 @@
...
+                    Maecenas quis turpis sit amet sapien dapibus sollicitudin viverra eu justo. Vivamus posuere metus sit amet purus tempus volutpat.
+                    Donec eleifend elit quam.
+                </p>
+                <!-- eslint-enable max-len -->
+            </EsCollapse>
+        </div>
+
+        <div class="mt-450 mb-450">
+            <h2>
+                Programmatic EsCollapse with Priority
```
ericdouglaspratt, line 82: Nit: We could axe the "EsCollapse" here since everything on this page is EsCollapse, and it could be something like "Full programmatic"

es-design-system/pages/molecules/es-collapse.vue
```diff
@@ -8,37 +8,118 @@
...
+            </h2>
             <p>
-                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi in aliquam ex. Nullam vestibulum ex mi, ut suscipit libero condimentum id.
-                Pellentesque eu diam vel nisi molestie porta eget sed odio. Quisque congue risus id metus facilisis, non imperdiet libero rutrum. Mauris
-                vitae ante porttitor, consectetur purus faucibus, euismod ex. Orci varius natoque penatibus et magnis dis parturient montes, nascetur
-                ridiculus mus. Nulla ullamcorper elit sed viverra finibus. Mauris vitae tortor mauris. Cras suscipit nibh nec nisi cursus ornare.
-                Maecenas quis turpis sit amet sapien dapibus sollicitudin viverra eu justo. Vivamus posuere metus sit amet purus tempus volutpat.
-                Donec eleifend elit quam.
+                An EsCollapse component that takes a "visible" prop with "is-programmatic-until-user-input" true. Click
+                the checkbox to toggle showing it's contents! Unlike the previous example, if you click the collapse
```
ericdouglaspratt, line 86: Typo: "it's" > "its"

es-design-system/pages/molecules/es-collapse.vue
```diff
@@ -8,37 +8,118 @@
...
+                <!-- eslint-enable max-len -->
+            </EsCollapse>
+        </div>
+
+        <div class="mt-450 mb-450">
+            <h2>
+                Programmatic EsCollapse
+            </h2>
+            <p>
+                An EsCollapse component that takes a "visible" prop. Click the checkbox to toggle showing it's
```
ericdouglaspratt, line 44: Typo: "it's" > "its"

es-design-system/pages/molecules/es-collapse.vue
```diff
@@ -8,37 +8,118 @@
...
+        <div class="mt-450 mb-450">
+            <h2>
+                Default EsCollapse
+            </h2>
+            <p>
+                A normal EsCollapse component. Click it to toggle showing it's contents!
+            </p>
+            <EsCollapse
+                id="defaultCollapse"
+                class="ml-450 mr-450">
```
ericdouglaspratt, line 20: If we add `mt-450` here, it'd better match the spacing of the other examples between description and collapse.

es-design-system/pages/molecules/es-collapse.vue
```diff
@@ -8,37 +8,118 @@
...
+
+        <div class="mt-450 mb-450">
+            <h2>
+                Programmatic EsCollapse
+            </h2>
+            <p>
+                An EsCollapse component that takes a "visible" prop. Click the checkbox to toggle showing it's
+                contents! If you click the collapse itself, the "visible" prop will no longer be honored.
+            </p>
+            <form class="ml-450 mr-450 mt-5">
```
ericdouglaspratt, line 47: `mt-5` is deprecated, we should replace with `mt-450`

es-design-system/pages/molecules/es-collapse.vue
```diff
@@ -8,37 +8,118 @@
...
-                ridiculus mus. Nulla ullamcorper elit sed viverra finibus. Mauris vitae tortor mauris. Cras suscipit nibh nec nisi cursus ornare.
-                Maecenas quis turpis sit amet sapien dapibus sollicitudin viverra eu justo. Vivamus posuere metus sit amet purus tempus volutpat.
-                Donec eleifend elit quam.
+                An EsCollapse component that takes a "visible" prop with "is-programmatic-until-user-input" true. Click
+                the checkbox to toggle showing it's contents! Unlike the previous example, if you click the collapse
+                itself, the "visible" prop will continue to be honored.
             </p>
-        <!-- eslint-enable max-len -->
-        </EsCollapse>
+            <form class="ml-450 mr-450">
```
ericdouglaspratt, line 89: If we add a `mt-450` here, it'd match the spacing of the other examples. Either that, or I could see aligning the checkbox with the description text above and keeping the collapse as the one that's indented.

ericdouglaspratt (approved):
Nice work! Very slick that it's backwards-compatible with Nathaniel's `visible` work but also works with `v-model` now.

A couple of optional comments on wording and spacing but otherwise good to go!

zacharyjarnagin:
Thanks for the suggestions @ericdouglaspratt! I addressed them all in my most recent commit

PR #1199, opened 2023-09-13
CSM-2859: Fixing Multiple File Download Bug

satwikgoyal:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CSM-2859

### ❓ Type of change

### 📚 Description

- Earlier when uploading a file one _after_ another in the same session would throw an error saying "You have already uploaded a file with the name "{fileName}". Please rename your file and try again.". This was caused because when the second file was being uploaded, we were re-uploading the first file causing issues when downloading these files.
- In `es-vue-base/src/lib-components/EsFileInput.vue` now uploading only `newValidFiles` instead of all `files`

### 🥼 Testing

- There are 3 ways we can test these changes:
   - Deploy this to production and then test out changes in `es-cdgm`
   - You'll need to symlink `~/Dev/es-ds/es-vue-base/dist` to `~/Dev/-es-cdgm/frontend/node_modules/@energysage/es-vue-base` (https://docs.npmjs.com/cli/v9/commands/npm-link)
   - Quick and dirty testing: copy the file in `es-cdgm` and import it from local components instead of `es-ds`
- Then upload files one _after_ another and try downloading them as a developer
- Try other ways to possibly cause these issues such as drag and drop or uploading multiple files at _once_, etc

### 📝 Checklist

### DIFF
--- a/es-vue-base/src/lib-components/EsFileInput.vue
+++ b/es-vue-base/src/lib-components/EsFileInput.vue
@@ -194,9 +194,9 @@ export default {
             this.files = this.files.filter(({ name }) => !newValidFiles.some((file) => file.name === name));
             this.files = [...this.files, ...newValidFiles];
 
-            if (this.files.length > 0) {
-                this.$emit('readyToUpload', this.files);
-                this.readFilesIntoUrl(this.files);
+            if (newValidFiles.length > 0) {
+                this.$emit('readyToUpload', newValidFiles);
+                this.readFilesIntoUrl(newValidFiles);
             }
             this.pickedItems = [];
         },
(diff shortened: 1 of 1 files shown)


zacharyjarnagin (commented):
I am not sure if we care about this use case but:

1. Upload a file
2. Upload a different file
3. Click submit

Only the last file gets uploaded. If that is okay, then this PR is good to go! Otherwise, we may need an additional map

zacharyjarnagin (approved):
After discussion, seems like my comment is the intended flow! Nice work

satwikgoyal:
> I am not sure if we care about this use case but:
> 
> 1. Upload a file
> 2. Upload a different file
> 3. Click submit
> 
> Only the last file gets uploaded. If that is okay, then this PR is good to go! Otherwise, we may need an additional map

For future developers: this case will be handled by the code.

satwikgoyal:
@zacharyjarnagin I found an edge case that I resolved today. As a combination to this branch I created another one in `es-cdgm` here: https://github.com/EnergySage/es-cdgm/pull/3044. Please lmk what you think about this.

zacharyjarnagin (approved):
Tested and verified working!! Nice job

PR #1255, opened 2023-12-06
feat: Improve phone number capture

nishamishra12:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/IDR-57

### ❓ Type of change

### 📚 Description

- Used https://vuejs-tips.github.io/vue-the-mask/ library to add masking to the phone number field
- The phone number field would be in use at multiple places, so I have conditionally added the mask with new type.
- Added the masked phone number field in molecules (FormInput) section as well as the form-field-validation section in examples
- Used this https://stackoverflow.com/questions/51836641/put-a-vue-directive-on-condition as a reference to conditionally use masking without null error
### 🥼 Testing

- Tested locally

-

### 📝 Checklist

### DIFF
--- a/es-vue-base/src/lib-components/EsFormInput.vue
+++ b/es-vue-base/src/lib-components/EsFormInput.vue
@@ -144,7 +144,7 @@ export default {
         },
         phoneMaskValue: {
             type: String,
-            default: '(###)-###-####',
+            default: '(###) ###-####',
             required: false,
         },
     },


es-vue-base/src/lib-components/EsFormInput.vue
```diff
@@ -130,6 +142,11 @@ export default {
             type: Boolean,
             default: false,
         },
+        phoneMaskValue: {
+            type: String,
+            default: '(###)-###-####',
```
nathanielwarner, line 147: This should be `(###) ###-####` rather than `(###)-###-####`

nathanielwarner (changes requested):
This looks generally good and will improve the user experience. There's one thing we should correct in order to have the right format for the phone number.

es-vue-base/src/lib-components/EsFormInput.vue
nishamishra12: Based on the screenshot that Xuewei  provided for design it is (###)-###-###. However, I can make this like the one you suggested as that is general US Format and it makes sense for the default to a standard format.

nathanielwarner (approved):
LGTM! 🚀

PR #1286, opened 2024-05-01
feat: new error page design

nathanielwarner:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1416

### ❓ Type of change

### 📚 Description

Updates the styling of the error page for ESDS 2.0 to match the Figma.

### 🥼 Testing

Manual testing

- General

### 📝 Checklist

### DIFF
--- a/es-vue-base/src/lib-components/EsErrorPage.vue
+++ b/es-vue-base/src/lib-components/EsErrorPage.vue
@@ -1,5 +1,5 @@
 <template>
-    <b-container class="py-450 bg-gray-50">
+    <b-container class="py-450 py-lg-800 bg-gray-50">
         <b-row>
             <b-col
                 md="7"
--- a/es-vue-base/src/lib-components/EsErrorPage.vue
+++ b/es-vue-base/src/lib-components/EsErrorPage.vue
@@ -1,5 +1,5 @@
 <template>
-    <b-container class="my-450 bg-gray-50">
+    <b-container class="py-450 bg-gray-50">
         <b-row>
             <b-col
                 md="7"
@@ -7,11 +7,11 @@
                 <h1 class="font-weight-bolder display-3 text-gray-600">
                     {{ errorShortMessage }}
                 </h1>
-                <h2
+                <p
                     id="msgError"
-                    class="h4 text-gray-600">
+                    class="font-size-lg text-gray-600">
                     {{ errorLongMessage }}
-                </h2>
+                </p>
             </b-col>
         </b-row>
         <b-row>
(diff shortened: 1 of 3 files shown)


ericdouglaspratt (changes requested):
Looks good!

Can we:
- Update all text to be sentence case rather than title case (e.g. "Access denied" instead of "Access Denied" - the design is incorrect here)
- Update the docs page to pass in text for the links that matches the design (except also in sentence case)
- Update the docs page to include an example for each supported error code
- Update the docs page with guidance on how to use this in a page layout (e.g. if we're adding the background color to the component, is this component sufficient to take over the entire viewport? or are we expecting them to also set the same background color on the page itself and set their own top padding of ~224px? if the former, we may want to increase the padding around the content to 100% match the design)

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -1,21 +1,16 @@
...
         <b-row>
             <b-col
                 md="7"
                 class="my-450 mx-auto text-center">
-                <h1 class="font-weight-boldest display-3">
-                    {{ errorType }}
+                <h1 class="font-weight-bolder display-3 text-gray-600">
+                    {{ errorShortMessage }}
                 </h1>
                 <h2
```
ericdouglaspratt, line 10: This was pre-existing, but semantically, this content should be a `<p>`, not an `<h2>` tag, as it's not a heading for content underneath it - can we update this as well

nathanielwarner:
> we may want to increase the padding around the content to 100% match the design)

@ericdouglaspratt To clarify, should I hardcode the pixel values for this? Using the `py-800` class gets us most of the way there but it's not 100% matching the design.

ericdouglaspratt:
I'd say whatever padding utility class is closest. Doesn't need to be exact.

es-vue-base/src/lib-components/EsErrorPage.vue
```diff
@@ -1,22 +1,17 @@
 <template>
-    <b-container class="my-450">
+    <b-container class="py-450 py-lg-800 bg-gray-50">
         <b-row>
             <b-col
                 md="7"
                 class="my-450 mx-auto text-center">
-                <h1 class="font-weight-boldest display-3">
-                    {{ errorType }}
+                <h1 class="font-weight-bolder display-3 text-gray-600">
```
ericdouglaspratt, line 7: Now that we have it, can we swap out `font-weight-bolder display-3` for `post1` class?

ericdouglaspratt (changes requested):
Just one small thing

es-vue-base/src/lib-components/EsErrorPage.vue
nathanielwarner: Done!

ericdouglaspratt (approved):
Nice!

PR #1318, opened 2024-05-02
feat: updated ESRating styling

jaissica:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-1419

### ❓ Type of change

### 📚 Description

- Updated ESRating styling - changed the color to orange. 
- Updated the star icon templates to match the figma design. 

### 🥼 Testing

- Tested the changes locally on http://localhost:8500/molecules/es-rating

- Should I update the Star icon (25% & 75%) on [icons page](http://localhost:8500/atoms/icons)?
- Anything I have missed?

### 📝 Checklist


ericdouglaspratt (changes requested):
I'm seeing the IconStarFull have the new rounded corners, but the IconStarEmpty and the IconStarHalf have sharp corners still - I think we need to fix the latter two.

### Figma
Screen Shot 2024-05-03 at 8 48 40 AM

### Code
Screen Shot 2024-05-03 at 8 48 52 AM

Also, this docs page does not sufficiently explain the capabilities of this component - can you add examples for the following static display states:
- zero stars
- 0.5 stars
- 1 star
- 1.5 stars
- 2 stars
- 2.5 stars
- 3 stars
- 3.5 stars
- 4.5 stars
- 5 stars

Also, as in the Figma, can you add examples of those above static display examples against the three non-white background colors, to show that they can appropriately render against darker backgrounds?
- gray 50
- blue 50
- blue 900

es-bs-base/scss/vue/components/_form-rating.scss
```diff
@@ -57,3 +57,9 @@
     }
   }
 }
+
+.rating {
+    display: flex;
+    height: 4.5rem;
+    width: 9.8rem;
```
ericdouglaspratt, line 64: It looks like this class is only used by the Rating docs page - if so, it shouldn't be part of the `es-bs-base` package, it should be a style in a `<style>` tag on that Rating docs page.

Also - you can essentially accomplish all of this using utility classes `d-flex p-200` on the div elements on the docs page itself, so you shouldn't need a `<style>` tag at all.

ericdouglaspratt (changes requested):
Looks good - just two items:

The Figma has the half star icon filled in with white in the empty part:
Screen Shot 2024-05-06 at 9 22 21 AM

Whereas in the code I see it as transparent:
Screen Shot 2024-05-06 at 9 22 14 AM

es-bs-base/scss/vue/components/_form-rating.scss
jaissica: Thanks for the feedback! I have updated the code as suggested.

ericdouglaspratt (changes requested):
~There appears to be an issue right now with how CI is configured to run when PRs are updated - we'll have to look into that - CI should fail on this PR to let you know that there's Jest snapshot issue at the moment.~ *Edit:* I think I just fixed it, it was only running on PRs against `main` rather than against the `esds-2.0-visual-identity` feature branch.

If you run `make lint && make test` locally in the `es-ds` root directory, you can see that since EsRating was updated, its Jest test snapshot is out of date and the test fails as a result.

Can you do `cd es-vue-base` and then run `npm run test:update` to update the Jest snapshots? And then `cd ..` to get back to the `es-ds` root directory and run `make lint && make test` again to confirm success (and/or push to the branch and CI should now run again)?

ericdouglaspratt (approved):
Nice! Thanks for catching the snapshot update for my EsViewMore changes as well - sorry about that.

PR #1320, opened 2024-05-03
CED-1417: EsPopover restyling

hroth1994:
### 🔗 Linked issue

* https://energysage.atlassian.net/browse/CED-1417

### ❓ Type of change

### 📚 Description

* Updated background and text colors, border radius, padding, close icon location, and arrow style based on figma design
* Updated docs to indicate which `placement` values to use based on figma design

### 🥼 Testing

* Tested the changes locally on http://localhost:8500/molecules/es-popover

- General feedback on docs and styling

### 📝 Checklist


es-bs-base/scss/_variables.scss
```diff
@@ -1511,26 +1511,26 @@ $form-feedback-tooltip-border-radius: $tooltip-border-radius !default;
 // Popovers
 
 $popover-font-size:                 $font-size-sm !default;
```
nathanielwarner, line 1513: The font size within the popover body should be 16, not 14. I think the one you want is `$font-size-100`.

es-bs-base/scss/_variables.scss
```diff
@@ -1511,26 +1511,26 @@ $form-feedback-tooltip-border-radius: $tooltip-border-radius !default;
...
-$popover-bg:                        $gray-900 !default;
+$popover-bg:                        $black !default;
```
nathanielwarner, line 1514: This is a pre-existing issue, but I noticed that if you use one of the [complicated examples from BootstrapVue](https://bootstrap-vue.org/docs/components/popover#advanced-b-popover-usage-with-reactive-content) for the default `b-popover`, the styling doesn't look right as a result of the dark theme being default. @ericdouglaspratt do you think it would make sense to change the default in the SCSS to the light theme?

Screenshot 2024-05-06 at 12 06 57

nathanielwarner (changes requested):
Other than the font size in the header, the font weight in the header should be 600 rather than 500. And in the dark theme, the link should be white rather than blue. (I agree the blue kind of looks better but the design has it as white.)

In addition to the question of whether or not we should be using dark theme by default, I'm wondering if this component has support for the orange button that's shown in the design. Again it might be a question for @ericdouglaspratt around scope, as the ticket does specify that functionality shouldn't be changed. If I put a button in right now, it's blue with a black background, which does have a contrast issue.

Screenshot 2024-05-06 at 12 23 35

hroth1994:
> Again it might be a question for @ericdouglaspratt around scope, as the ticket does specify that functionality shouldn't be changed. Also noting, If I put a button in right now in the light mode, it's blue with a black background, which does have a contrast issue.

I updated the link to be white in dark mode and am leaving the button styling to the EsButton component

es-bs-base/scss/_popover.scss
```diff
@@ -24,6 +24,7 @@
   border: variables.$popover-border-width solid variables.$popover-border-color;
   @include border-radius.border-radius(variables.$popover-border-radius);
   @include box-shadow.box-shadow(variables.$popover-box-shadow);
```
ericdouglaspratt, line 26: Noticed this:
Screen Shot 2024-05-07 at 11 45 43 AM

Had a [conversation with Lauren in Slack](https://energysage.slack.com/archives/C02FU6TCDPX/p1715097269775819) about taking the Modal shadow and using it on both the dark and light versions of the Popover and she agreed - we may tweak the shadow later, but can you take the one from Modal in the Figma and put it in the `$popover-box-shadow` variable? You may need to remove the wrapping `box-shadow.box-shadow()` function since that won't print it out since shadows are disabled globally (and should probably stay that way, lest we affect everything).

Then it'll look more like this, which is definitely better for visibility, though maybe design will tweak the color values later on:
Screen Shot 2024-05-07 at 11 52 55 AM

es-design-system/pages/molecules/es-popover.vue
```diff
@@ -21,16 +21,22 @@
                 </b-link>
                 for more details.
             </p>
-            <div class="my-450">
```
ericdouglaspratt, line 24: We generally do have `my-450` around each category of example(s) on a page to give it room and to visually separate the examples - I'd recommend keeping that around each variant here.

es-design-system/pages/molecules/es-popover.vue
```diff
@@ -39,7 +45,8 @@
                         </b-link>
                     </label>
                     <EsPopover
-                        target="darkTitleTarget"
+                        :placement="placement.directive"
+                        :target="`darkTitleTarget-${placement.directive}`"
                         variant="dark">
                         <template #title>
                             My Title
```
ericdouglaspratt, line 52: Nit: I realize this was pre-existing on this page but all titles, all links, everywhere, should always be sentence case rather than title case - I'd like us to work towards making the design system consistent with its own guidance on this, otherwise people may copy and paste this example and do the wrong thing - can we update this to "My title", the links to "Learn more", etc. throughout this page?

es-vue-base/src/lib-components/EsPopover.vue
```diff
@@ -15,6 +15,7 @@
             -->
             <slot name="title" />
             <EsButton
+                :inline="true"
```
ericdouglaspratt, line 18: Nit: for brevity, this could just be `inline` rather than `:inline="true"`. Vue generally matches the rules for HTML tag attributes (e.g. `<EsButton>` should really be `<es-button>` but we should really have a linting rule for that like we do on our other repos).

es-vue-base/src/lib-components/EsPopover.vue
```diff
@@ -32,8 +33,9 @@
                 <slot />
                 <div v-if="!hasTitle">
                     <EsButton
+                        :inline="true"
                         variant="link"
-                        class="p-0"
+                        class="p-0 pl-2"
```
ericdouglaspratt, line 38: Per the [design system's spacing page](https://design.energysage.dev/atoms/spacing), `pl-2` is deprecated and should be `pl-50`.

ericdouglaspratt (changes requested):

es-vue-base/src/lib-components/EsPopover.vue
```diff
@@ -212,14 +214,9 @@ export default {
     }
 
     .popover-body {
-        .btn, .btn:hover, .btn:active {
+        .btn, .btn:hover, .btn:active, a {
             color: variables.$white;
```
ericdouglaspratt, line 218: For buttons, this color happens to be correct for the `dark-bg` variant, but it prevents that variant's outline version from getting the correct non-white hover and active colors. I think we should remove all the `.btn` classes from this rule and just leave `a`.

Screen Shot 2024-05-07 at 5 01 39 PM

es-bs-base/scss/_variables.scss
ericdouglaspratt: I sincerely hope we're not putting forms in popovers anywhere, because popovers are meant to be extremely temporary and can too easily be accidentally dismissed and you lose data that you've entered.

As the default page background is white, it makes sense to me that most cases of the popover should be black, and therefore the default should be black.

ericdouglaspratt (changes requested):
I found a slight issue on mobile with the "right bottom" example completely out of position - anything we can do, or is that just the buggy behavior of the Bootstrap Vue tooltip when you have that setting and there isn't enough room on the right? I'm guessing the latter, and that it's only smart enough to say "I don't have enough room on the right, so I must try to appear on the left side" which it's clearly trying to do but it has even less room on the left side so it's being pushed over to the right side weirdly, but I figured I'd ask.

Screen Shot 2024-05-07 at 4 40 43 PM

nathanielwarner (approved):
Nice work on this!

PR #1321, opened 2024-05-03
CED-1418: pagination style update

nathanielwarner:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1418

### ❓ Type of change

### 📚 Description

Implements the new style for EsPagination as per the Figma

### 🥼 Testing

Manual testing

- Does the display on mobile devices look reasonable? I'm not thrilled about allowing the ellipses to be removed, but I'm not sure there's a better solution given how large we want the buttons to be now
- I tried to make the changes in the SCSS files where appropriate, knowing that this affects how it looks if you use `b-pagination` directly. I temporarily put that component as well on the docs page to verify that nothing crazy happened. Do you think I struck the right balance here, or would it be better to apply more of the changes to the SCSS file? Fewer?

### 📝 Checklist


ericdouglaspratt (changes requested):
I think we _should_ affect BPagination with these styles - we want to ensure that all UI (if it's a light lift) is affected by viz id - so I'd say feel free to move them into that _pagination.scss and give it a test to make sure it works.

And yeah I played around with a few `:nth-child` and next-sibling and subsequent-sibling combinators to try to get ellipses but not too many numbers from showing up on mobile, and it worked nicely for many-page lists, but not for smaller lists, so I think this is the best we can do for now.

nathanielwarner:
@ericdouglaspratt I've applied the styles to the SCSS where it was a low lift, and validated it by temporarily putting the stock `b-pagination` components in the docs page. In particular, stock usages of `b-pagination` will now have the same colors as our custom component for the most part. But there's some discrepancies that are hard to fix since the stock component doesn't distinguish between normal links and the left/right arrows, or a disabled link and the ellipse.

ericdouglaspratt (approved):
Looks good! And yeah bootstrap vue pagination has enough extra going on that's fine.

PR #1325, opened 2024-05-06
CED-1420: EsSlider restyling

hroth1994:
### 🔗 Linked issue

* (https://energysage.atlassian.net/browse/CED-1420)

### ❓ Type of change

### 📚 Description

* Updated marks, slider, slider dot, and tooltip styles based on figma design
* Involved a refactor to move styles from `es-vue-base` into `es-bs-base`

### 🥼 Testing

* Tested the changes locally on http://localhost:8500/molecules/es-slider

* General feedback on styling

### 📝 Checklist


es-vue-base/src/lib-components/EsSlider.vue
```diff
@@ -17,15 +17,15 @@
             <!-- Tooltip above the slider thumb -->
             <template #tooltip="{ value, focus }">
                 <div
-                    class="slider-tooltip d-flex h5 align-items-center text-white justify-content-center m-0 bg-primary"
+                    class="slider-tooltip d-flex h4 align-items-center text-white justify-content-center m-0 bg-blue-600 font-weight-boldest"
```
ericdouglaspratt, line 20: Interesting that this was an `h5` before - I believe `h4`, though, makes it larger than the design on desktop (if you drag the slider all the way to the right, the "$1000" doesn't fit inside the circle and the font size is 20px rather than the 16px in the design) - maybe we just remove the class altogether and let it take the default 16px body font size on both mobile and desktop?

es-bs-base/scss/vue/components/_slider.scss
```diff
@@ -0,0 +1,49 @@
...
+
+  .vue-slider-rail {
+    height: variables.$b-slider-rail-height;
+  }
+
+  .vue-slider-dot-handle {
+    border: variables.$b-slider-dot-border;
+  }
+
+  .slider-tooltip {
```
ericdouglaspratt, line 27: According to the Figma design, I believe the tooltip circle should also have a box shadow on it.

es-bs-base/scss/vue/components/_slider.scss
```diff
@@ -0,0 +1,49 @@
...
+
+  .vue-slider-process {
+    background: variables.$b-slider-background;
+  }
+
+  .vue-slider-rail {
+    height: variables.$b-slider-rail-height;
+  }
+
+  .vue-slider-dot-handle {
```
ericdouglaspratt, line 23: It looks like the Figma design specifies a different shadow than the dot handle is getting right now - maybe we can put that one here.

es-vue-base/src/lib-components/EsSlider.vue
```diff
@@ -17,15 +17,15 @@
...
                     :class="[{ focus }]">
                     {{ tooltipFormatter(value) }}
                 </div>
             </template>
             <!-- Labels below the process bar -->
             <template #label="{ label, active }">
                 <div :class="['slider-label', 'vue-slider-mark-label', { active }]">
-                    {{ label }}
+                    ${{ label }}
```
ericdouglaspratt, line 28: This slider may be used in cases that don't involve dollar amounts, so I'm not sure we should hardcode a "$" in front of each label - does the library provide any label formatter option, similar to what we're doing with tooltip formatter?

ericdouglaspratt (changes requested):
Nice! A couple items, plus I believe I see some CI failures at the bottom of this PR that need to be fixed.

ericdouglaspratt (approved):
Looks good!

PR #1326, opened 2024-05-06
CED-1444: horizontal list

nathanielwarner:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1444

### ❓ Type of change

### 📚 Description

Updated `EsHorizontalList` to the ESDS 2.0 theme

### 🥼 Testing

manual testing

- There's a discrepancy in the figma, as it says "Disabled icon is grey 500", but it's actually displaying gray 600. I'm using gray 500 since that's explicitly stated, and matches what's used in the pagination component.
- This component is custom-designed, and is not accessible at all. I suppose it's outside the scope of this PR to fix, but what do you think about adding a note to the documentation of this component that using it should be avoided due to accessibility? In our repos, I see it's used for `ArticleRelated` currently.

### 📝 Checklist

### DIFF
--- a/es-vue-base/src/lib-components/EsHorizontalList.vue
+++ b/es-vue-base/src/lib-components/EsHorizontalList.vue
@@ -28,7 +28,11 @@
             <es-button
                 v-if="arrows"
                 variant="link"
-                class="text-black d-none d-lg-block p-0 mr-200"
+                :disabled="atBeginning"
+                class="d-none d-lg-block p-0 mr-200"
+                :class="{
+                    'text-gray-900': !atBeginning,
+                }"
                 @click="arrowClick('left')">
                 <icon-chevron-left />
                 <span class="sr-only">Previous</span>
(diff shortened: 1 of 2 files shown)


es-vue-base/src/lib-components/EsHorizontalList.vue
```diff
@@ -28,7 +28,11 @@
             <es-button
                 v-if="arrows"
                 variant="link"
-                class="text-black d-none d-lg-block p-0 mr-200"
+                :disabled="atBeginning"
+                class="d-none d-lg-block p-0 mr-200"
```
ericdouglaspratt, line 32: The design has the chevrons a bit closer to the circles. I doubt this design passes the tap-targets accessibility test as those circles and chevrons are pretty close together, but that part is out of scope - to better match the design without making the chevrons too close, maybe we bump this `mr-200` down to `mr-100`?

es-vue-base/src/lib-components/EsHorizontalList.vue
```diff
@@ -46,7 +50,11 @@
             <es-button
                 v-if="arrows"
                 variant="link"
-                class="text-black d-none d-lg-block p-0 ml-200"
```
ericdouglaspratt, line 49: Similar here - maybe `ml-100`?

ericdouglaspratt (changes requested):
Looks good - two small tweaks.

es-vue-base/src/lib-components/EsHorizontalList.vue
nathanielwarner: Ah yes good catch!

ericdouglaspratt (approved):
Nice!

PR #1354, opened 2024-05-21
CED-1509: fix styling of ReviewsIoCarousel in solar cost/companies

hroth1994:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1509

### ❓ Type of change

### 📚 Description

- The top of the review cards was being cut off when there was no header - adding `margin-top: 0 !important;` to `R-ReviewsList` fixed this

### 🥼 Testing

- Tested on mobile and desktop locally on the home page, solar cost page, solar companies page, and `es-ds`

- General

### 📝 Checklist


ericdouglaspratt (changes requested):
Hmm that does work to prevent it from getting cut off, but it also means the spacing between this reviews carousel and what's above it on the page (as set by the above block's margin bottom) will be 5px fewer than it should be, because their styles are doing this `margin-top: -5px`, which moves the entire thing up 5px, and our removal of `overflow: hidden` allows that to move up 5px before when the component should actually start.

Screen Shot 2024-05-22 at 4 28 18 PM

I might instead recommend adding a `margin-top: 0` in our `.R-ReviewsList` class in this file to cancel out that `margin-top: -5px` (may or may not need to add parent classes around it to increase the specificity enough to override the above class coming from them, but try it normally first and see if that works).

ericdouglaspratt (approved):
Looks good! Tested via doing the same changes manually to the solar cost page in dev tools.

PR #1384, opened 2024-06-10
feat: state icons

nathanielwarner:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1567

### ❓ Type of change

### 📚 Description

Add SVG icons for every state, plus DC, minus AK and HI. Copied from local-data and modified to provide sizing

### 🥼 Testing

Manual testing on http://localhost:8500/atoms/icons

- General

### 📝 Checklist


es-vue-base/src/lib-icons/states/icon-state-dc.vue
```diff
@@ -0,0 +1,43 @@
...
+            fill="currentColor"
```
ericdouglaspratt, line 11: So DC is the only state graphic that had two colors before, per Maria's design, to highlight the DC part of Maryland. ([example rebates & incentives page](https://www.energysage.com/local-data/solar-rebates-incentives/dc/))

Can you check with Asheeta on how important she feels it is to keep those two colors? If so, maybe we default to a certain darker orange for the inner DC piece and accept a prop for that secondary color that can be passed in to customize that default?

Screen Shot 2024-06-11 at 6 45 50 AM

ericdouglaspratt (changes requested):
Looks good - just one item.

es-vue-base/src/lib-icons/states/icon-state-dc.vue
nathanielwarner: I assumed that was a mistake, and didn't even notice the DC cutout 😂. I'll check with Asheeta

nathanielwarner: Ok, I've made it so the surrounding area (not DC) is brighter than the rest using a CSS filter. I figure this will be a saner default value than a hardcoded orange. The brightness is also adjustable by prop.

ericdouglaspratt (approved):
Nice!

PR #1400, opened 2024-06-18
feat: add autofocus to the first zip entry on every page

LiamKosar:
### 🔗 Linked issue
- https://energysage.atlassian.net/browse/GROW-350

### ❓ Type of change

### 📚 Description

- Bound the autofocus attribute of the es-form-input component WITHIN EsZipCodeForm.vue to the autofocus data element. Default value is false (no autofocus), but if the EsZipCodeForm is the first EsZipCodeForm on the page, the mounted function will change the value to true, autofocusing the first EsZipCodeForm on every page using the es-ds system.
- Changed the id's of the EsZipCodeForms within the es-zip-code-form.vue page to be unique, otherwise the code won't work properly. Storyblok creates unique id's for every EsZipCodeForm, so this should not be an issue on our real pages.

- https://energysage.atlassian.net/browse/GROW-350

### 🥼 Testing

- Tested locally at http://localhost:8500/organisms/es-zip-code-form

-

### 📝 Checklist

### DIFF
--- a/es-vue-base/src/lib-components/EsZipCodeForm.vue
+++ b/es-vue-base/src/lib-components/EsZipCodeForm.vue
@@ -173,14 +173,6 @@ export default {
             zipCode: this.zipCodeValue,
         };
     },
-    mounted() {
-        let inputField = document.getElementsByName('zip_code')[0];
-        if(this.inputId === inputField.id) {
-            inputField.focus({
-                preventScroll: true
-            });
-        }
-    },
     computed: {
         stackBreak() {
             let { stackUntil } = this;
@@ -194,6 +186,14 @@ export default {
         zipCodeValue(newVal) {
             this.zipCode = newVal;
         },
+    },
+    mounted() {
+        const inputField = document.getElementsByName('zip_code')[0];
+        if(this.inputId === inputField.id) {
+            inputField.focus({
+                preventScroll: true
+            });
+        }
     },
     methods: {
         handleSubmit() {
--- a/es-vue-base/src/lib-components/EsZipCodeForm.vue
+++ b/es-vue-base/src/lib-components/EsZipCodeForm.vue
@@ -24,7 +24,6 @@
                 <es-form-input
                     :id="inputId"
                     v-model="zipCode"
-                    :autofocus="autofocus"
                     autocomplete="postal-code"
                     class="mb-100"
                     :class="{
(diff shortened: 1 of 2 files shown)


es-vue-base/src/lib-components/EsZipCodeForm.vue
```diff
@@ -175,8 +172,16 @@ export default {
     data() {
         return {
             zipCode: this.zipCodeValue,
+            autofocus: false,
         };
     },
+    mounted() {
+        if(this.inputId === document.getElementsByName('zip_code')[0].id) {
+            console.log("This id: " + this.inputId);
+            console.log("First id: " + document.getElementsByName('zip_code')[0].id);
```
nathanielwarner, line 181: Looks like console logs were left in

es-vue-base/src/lib-components/EsZipCodeForm.vue
```diff
@@ -86,9 +85,7 @@
                         class="text-nowrap"
                         :class="dark ? 'text-white' : ''"
                         :target="privacyPolicyNewTab ? '_blank' : '_self'">
-                        <slot name="privacyPolicyLinkText">
-                            Privacy Policy
-                        </slot>
+                        <slot name="privacyPolicyLinkText"> Privacy Policy </slot>
```
nathanielwarner, line 88: We should remove extraneous stylistic changes, especially since it causes this warning

es-vue-base/src/lib-components/EsZipCodeForm.vue
```diff
@@ -24,6 +24,7 @@
                 <es-form-input
                     :id="inputId"
                     v-model="zipCode"
+                    :autofocus="autofocus"
```
nathanielwarner, line 27: Noting the Github warning below. Using autofocus [causes screen readers to "teleport"](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus) to the zip code form on page load, which is an accessibility failure. I confirmed this behavior with VoiceOver. This issue actually affects our homepage as well due to it being used there, and we probably need to have a broader conversation about it. @ericdouglaspratt what do you think?

LiamKosar: Problem solved by using the html element method "focus" which allows me to pass in the prop "preventScroll: true"

nathanielwarner: This unfortunately doesn't solve the problem. With these changes, when you navigate to the page, the keyboard focus changes to the zip code form. This is a problem for people with vision impairment who rely on screen readers, because instead of starting at the top of the page, the screen reader will "teleport" to the zip code form.

nathanielwarner:
As I'll be out for the rest of the week, I need to hand this review off to someone else. The accessibility issue seems to be fundamental to what we're trying to do in this PR. But otherwise the code looks good to me.

ericdouglaspratt (commented):
@LiamKosar I think we can close this PR, due to Luke's decision not to roll out this test anymore?

PR #1461, opened 2024-07-24
feat: ESDS 3.0 release based on Vue 3 and PrimeVue

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-1339

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

- New `es-ds-components` package that is a Nuxt layer package and contains the PrimeVue config for downstream Nuxt 3 applications, as well as all related components
- New `es-ds-docs` folder that contains a Nuxt 3 app to power the new design system documentation site
- Local development and hot reloading are now _much_ faster and simpler; the necessary linking and dev server initialization can be accomplished with a single command: `make dev`

### 🥼 Testing

- Tested on the new local docs site

#### 🧐 Feedback Requested / Focus Areas

- Overall

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.
- [x] I have documented testing approach

PR #1467, opened 2024-08-07
CED-1724: new folder structure and publish process for v3

nathanielwarner:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1724

### ❓ Type of change

### 📚 Description

- Removes all legacy stuff (es-design-system and es-vue-base)
  - Will be retained along with the old es-bs-base on a separate fork before merging the Vue 3 branch into main
- Renames es-bs-base to es-ds-styles and removes JavaScript stuff from it
- Sets up a publishing process for v3. Rather than using lerna or another solution, it's just managed through the makefile. `npm publish` is invoked for `es-bs-base` (using the alpha tag) and `es-ds-components`
  - Management of version numbers is therefore manual for the time being.
  - The new steps to publish is documented in the readme
- Switch to explicit imports of PrimeVue components to enable tree shaking
- Move the nuxt config from es-ds-docs to es-ds-components. Ditto for package json dependencies

### 🥼 Testing

Manually tested publish process

- My update-docs-deps section of the makefile currently uninstalls and reinstalls `es-ds-styles` and `es-ds-components` in order to get the latest version. My attempts with using `npm update` were unsuccessful. Do you have any pointers here?

### 📝 Checklist


es-ds-docs/package.json
lgeggleston, line 1: Does `es-ds-docs` need its own version defined as well or not in this case?

Makefile
```diff
@@ -25,34 +28,58 @@ unlink:
 
 .PHONY: lint
 lint:
+	npm --prefix es-bs-base run lint
+# TODO: set up linting for es-ds-components and es-ds-docs
```
lgeggleston, line 32: is this remaining linting captured in a ticket somewhere?

Makefile
```diff
@@ -25,34 +28,58 @@ unlink:
...
+	npm --prefix es-ds-docs run build
+# es-ds-components does not have a build step
+
+.PHONY: legacy-build
+legacy-build:
 	npx lerna run build
 
 .PHONY: publish
 publish:
+	cd es-bs-base; npm publish --tag alpha
```
lgeggleston, line 63: I see the test 3.0 alpha release in `npm` but not seeing it in the github release [tags](https://github.com/EnergySage/es-ds/tags).  Should it be visible there as well?

lgeggleston (commented):
Your updates to make all the 2.0 processes `legacy` and to the package files make sense; README updates also look good! The updated Makefile commands make sense too I think, though I have limited experience with them so curious for Tom & Eric's thoughts when they're back.  
General question, though looks like it works fine, what are the benefits/considerations to moving away from Lerna or a different monorepo manager?

Makefile
nathanielwarner: [Yup](https://energysage.atlassian.net/browse/CED-1726)! Not sure of the status on that one

es-ds-docs/package.json
nathanielwarner: `es-ds-docs` isn't gonna be published as a package, so I will hesitantly say "no"

Makefile
nathanielwarner: At the moment, no. I tried to do the absolute minimum necessary to get the NPM publishing working. I'm not entirely clear on whether there's any purpose to the GitHub releases on this repo, but also probably a Tom or Eric question.

nathanielwarner:
> curious for Tom & Eric's thoughts when they're back.

Agreed! No rush in any case.

> what are the benefits/considerations to moving away from Lerna or a different monorepo manager?

For the time being, I didn't mess with lerna because I didn't want to break the legacy build and publishing which does rely on lerna. And to be honest, I'm also not sure what value lerna adds compared to just having the sequence of steps in the makefile. Maybe the automatic version increment comes from there? This is definitely something for Tom and Eric to weigh in on.

I mentioned it in [the big PR](https://github.com/EnergySage/es-ds/pull/1461#pullrequestreview-2219596737) as well, but it seems like it'd be easier to shed the legacy stuff in this branch. The old (v2) design system could be maintained on its own branch after merging v3.

Makefile
```diff
@@ -25,34 +28,58 @@ unlink:
 
 .PHONY: lint
 lint:
+	npm --prefix es-bs-base run lint
+# TODO: set up linting for es-ds-components and es-ds-docs
+# 	npm --prefix es-ds-components run lint
+# 	npm --prefix es-ds-docs run lint
+
+.PHONY: legacy-lint
+legacy-lint:
```
tomleo, line 37: I wonder if we should do this the other way around i.e. `lint` runs the `lint` command for all packages, and `lint-next` runs only prime-related packages. This would avoid having to change anything in regular CI/CD as we're still maintaining all packages ATM (i.e. `es-vue-base`)

Makefile
```diff
@@ -25,34 +28,58 @@ unlink:
...
 	npx lerna run lint
 
 .PHONY: test
 test:
```
tomleo, line 41: Similar to above, I might suggest we make the `test` command test everything, whereas perhaps a `test-next` is scoped to newer/prime related packages.

README.md
```diff
@@ -262,29 +262,48 @@ in the changelog when publishing a new release.
 For simplicity of deployment, versioning of packages are fixed and updated
 together.
 
-#### Publishing a new version of a package
+#### Publishing a new version of a package (3.0 and later)
```
tomleo, line 265: So long as we maintain all the packages in this repo, I think directions should include how to lint, test, and deploy all packages in this repo.

Makefile
```diff
@@ -1,78 +1,60 @@
-# Run local v3 docs site with hot reloading hooked up to es-bs-base and es-ds-components
+# Run local v3 docs site with hot reloading hooked up to es-ds-styles and es-ds-components
 
 .PHONY: dev
-dev:
-	cd es-bs-base; npm link
-	cd es-ds-components; npm link
-	cd es-ds-docs; npm link @energysage/es-bs-base @energysage/es-ds-components
+dev: symlink
```
tomleo, line 4: Nice 👍

Makefile
```diff
@@ -1,78 +1,60 @@
...
 
 .PHONY: build
 build:
-	npx lerna run build
+	npm --prefix es-ds-styles run build
+	npm --prefix es-ds-docs run build
+# es-ds-components does not have a build step
 
 .PHONY: publish
 publish:
```
tomleo, line 44: 🤔 If we're not going to use lerna, I wonder if we should remove this from the makefile. Using something like release-please (like in `es-storyblok-shared` might be a better solution).

What do you think about removing the `publish` command for now?

es-ds-docs/package.json
```diff
@@ -10,8 +10,8 @@
     "postinstall": "nuxt prepare"
   },
   "dependencies": {
-    "@energysage/es-bs-base": "^2.0.4",
-    "@energysage/es-ds-components": "^0.0.1",
+    "@energysage/es-ds-components": "^3.0.0-alpha.6",
```
tomleo, line 13: We might want to consider reverting this to `^0.0.1` since it's independent of the previous packages (same with `es-ds-styles`)

tomleo (approved):
A few things inline, also couldn't find good documentation around tsconfig's in mono-repos 🤔, but otherwise 👍

.github/workflows/ci.yml
```diff
@@ -40,10 +40,6 @@ jobs:
             - name: Install dependencies 👨🏻‍💻
               run: |
                 make install
-                make update-peer-deps
-                make build-scss-pkg
-                make symlink
-                make build-vue-pkg
                 make symlink
```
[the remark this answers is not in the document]
nathanielwarner: I would think we still need it for the same reason we did before... so that the versions of things in the repo are used, rather than the currently published versions.

Makefile
nathanielwarner: Yeah that makes sense. All done!

es-ds-docs/package.json
nathanielwarner: The one catch with that is that we would then be unable to ever have a release of es-ds-components or es-ds-styles with these versions that are already published because of [a restriction of NPM](https://docs.npmjs.com/cli/v10/commands/npm-unpublish). In any case we can certainly decide to rename things or accept that limitation if we prefer. Maybe worth a followup conversation

PR #1468, opened 2024-08-08
feat: globally style PrimeVue radio buttons

hroth1994:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1753

### ❓ Type of change

### 📚 Description

Globally styles PrimeVue radio buttons to match the [custom styling done for viz id on BFormGroup and BFormRadioGroup radio buttons](https://design.energysage.dev/molecules/radio-button) and created a “Radio button” molecule docs page 

### 🥼 Testing

Tested inline, stacked, and disabled radio button groups with focus and hover states locally at http://localhost:8500/molecules/radio-button

- I attempted to avoid regressions by adding new classes, is the resulting css too complicated? In addition, `has` is a relatively [new css feature](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) and may not be supported by older devices or browsers.
- `PrimeVue RadioButton` is a tad finicky and will not work if any props vary from the example, should a component `EsRadioButton` be created to deal with the classes and props necessary to get `PrimeVue RadioButton` working as expected?
- Should `ariaLabel` be added?

### 📝 Checklist

### DIFF
--- a/es-ds-docs/pages/molecules/radio-button.vue
+++ b/es-ds-docs/pages/molecules/radio-button.vue
@@ -18,18 +18,13 @@
             <p>
                 Please choose your favorite fruit.
             </p>
-            <div>
-                <div v-for="fruit in fruits" class="custom-control custom-control-inline custom-radio">
-                    <radio-button
-                        class="custom-control-input"
-                        input-class="custom-radio-input"
-                        v-model="selectedFruit"
-                        :input-id="`${fruit.key}-inline`"
-                        name="inline"
-                        :value="fruit.key" />
-                    <label :for="`${fruit.key}-inline`" class="custom-control-label">{{ fruit.name }}</label>
-                </div>
-            </div>
+            <es-radio-button
+                v-for="fruit in fruits"
+                v-model="selectedFruit"
+                :display-name="fruit.name"
+                group-name="inline"
+                :value="fruit.key"
+                inline />
         </div>
 
         <div class="my-500">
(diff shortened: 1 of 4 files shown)


es-ds-docs/pages/molecules/radio-button.vue
```diff
@@ -0,0 +1,105 @@
...
+            </p>
+            <div>
+                <div v-for="fruit in fruits" class="custom-control custom-control-inline custom-radio">
+                    <RadioButton
+                        class="custom-control-input"
+                        inputClass="custom-radio-input"
+                        v-model="selectedFruit"
+                        :inputId="`${fruit.key}-inline`"
+                        name="inline"
+                        :value="fruit.name" />
```
nathanielwarner, line 29: The `value` for each radio button should correspond to the "key" of each fruit (so this should be `:value="fruit.key"`). Right now, setting `selectedFruit` does not do anything because of the mismatch.

es-ds-docs/pages/molecules/radio-button.vue
```diff
@@ -0,0 +1,105 @@
...
+        <ds-doc-source
+            :doc-code="docCode"
+            doc-source="es-ds-docs/pages/molecules/radio-button.vue" />
+    </div>
+</template>
+
+<script setup lang="ts">
+import { ref } from 'vue';
+
+const selectedFruit = ref('');
```
nathanielwarner, line 87: To match the current design system, this should be "banana"

nathanielwarner (changes requested):
Nice work on this! There's a couple things to address that I mentioned inline.

> I attempted to avoid regressions by adding new classes, is the resulting css too complicated? In addition, has is a relatively [new css feature](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) and may not be supported by older devices or browsers.

I don't see any issue with this. The CSS isn't that complex, and `has` is supported by all the evergreen browsers.

> PrimeVue RadioButton is a tad finicky and will not work if any props vary from the example, should a component EsRadioButton be created to deal with the classes and props necessary to get PrimeVue RadioButton working as expected?

Could you elaborate on how it's finicky? I agree it might be better to create a wrapper component to make things easier. I commented on [the checkbox ticket](https://energysage.atlassian.net/browse/CED-1741) as well that I think we should consider maintaining the current API. At the very least, it doesn't seem like we should make downstream users deal with adding the `custom-control custom-control-inline custom-radio custom-control-input custom-radio-input` classes.

> Should ariaLabel be added?

What would we be adding `aria-label` to? The radio buttons already have labels associated with them.

nathanielwarner (approved):
LGTM! 🚀

PR #1477, opened 2024-08-14
CED-1737: EsAccordion

nathanielwarner:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1737

### ❓ Type of change

### 📚 Description

- Implement EsAccordion v3 using PrimeVue. This is a wrapper component around using PrimeVue `Accordion` and `AccordionTab`.

### 🥼 Testing

Manually tested on http://localhost:8500/molecules/accordion

- General
- I have sunk many hours at this point into trying to make the collapse and expand functionality work smoothly. At the moment, the max height for the animation is hardcoded to 100px, which does work but results in a "jump" when the height of the inner element is bigger. I noticed that the implementation of EsCollapse currently has the size hardcoded to 3000px which has a similar downside. PrimeVue's [tailwind preset](https://tailwind.primevue.org/accordion/) uses 1000px. I'm at a bit of a loss, and at this point have gone beyond the expected story points. Things I've tried:
  - Dynamically set the max height based on the true height of the element, and relay that as a CSS variable for the transition. Problems:
    - For reasons I don't understand, the animation does not work if it's based on a CSS variable that is not hardcoded. As soon as the value of the variable changes, the transition is no longer animated.
    - I also have not had success getting the height of each collapse content. You can use scrollHeight after load, but that only works for the ones that are visible; other return 0. Same issue trying to do it on mount. I even tried a crazy workaround of rendering a clone of the element far offscreen but technically visible. That didn't work.. for some reason the height was always 60. Maybe because the chain of CSS classes is not there so things are not the right size. I noticed that BootstrapVue uses getBoundingClientRect instead so that might work, but it doesn't matter anyways because of the other limitation.
  - Base the animation on `transform: translateY` rather than `max-height`. With some other edits to the CSS this does actually do *almost* what we want, but the bounding box of the parent element does not respond to the transformation (which it turns out is an inherent limitation with transforms), so while the text slides in and away properly, the size of the collapse jumps in and out only at the end of the animation.
  - Try copying and pasting what BootstrapVue does. I haven't had any success with this- it just jumps immediately open and closed. I've shelved this code, and could potentially make it work with more fiddling but I'm not really sure.
- Maybe the best/only solution is to find a better hardcoded value? It's annoying because BootstrapVue does this properly.

### 📝 Checklist


es-ds-docs/pages/molecules/accordion.vue
```diff
@@ -0,0 +1,499 @@
...
+        Defaults to 'default'; also accepts 'minimal'. Default has a rounded white background,
+        minimal has no background, Heading 3 text style, reduced padding, and a horizontal border
+        separating accordions.
+        `,
+    ],
+];
+const accordionProps = [
+    [
+        'id',
+        '',
```
hroth1994, line 34: Should we put n/a for props with no default value? I saw it done in at least 1 other place

hroth1994 (commented):
I put a few initial comments - I'm going to take a deeper look at the CSS for the jumping animation

es-ds-docs/pages/molecules/accordion.vue
```diff
@@ -0,0 +1,499 @@
...
+    $prism.highlight();
+}
+</script>
+
+<template>
+    <div>
+        <h1>Accordion</h1>
+        <p class="mb-450">
+            Extended from
+            <a href="https://primevue.org/accordion/">PrimeVue accordion</a>
```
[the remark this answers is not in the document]
nathanielwarner: Very good point!

es-ds-components/components/es-accordion-list.vue
```diff
@@ -0,0 +1,112 @@
...
+const activeIndex = ref();
+
+watch(model, (newVal, oldVal) => {
+    if (newVal) {
+        activeIndex.value = children.findIndex((child) => child.props.id === newVal);
+    }
+})
+
+const accordionTabs = children.map((child, index) => {
+    if (child.props.id === props.initialExpandedId) {
```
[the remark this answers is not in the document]
nathanielwarner: Haha yes good point

es-ds-docs/pages/molecules/accordion.vue
nathanielwarner: I'm not sure.. does that run the risk of people thinking that "N/A" is the default value? I figure a blank space means there's no default value

nathanielwarner: Looking at this with a fresh set of eyes I think you're right, and it's no big deal in any case

hroth1994 (approved):
This looks great! Nice work 🌟

PR #1479, opened 2024-08-15
feat: EsDataTableSimple molecule for es-ds 3.0

jaissica:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-1743

### ❓ Type of change

### 📚 Description

- Created data-table-simple molecule for es-ds 3.0
- Created es-data-table-simple component based on es-ds 2.0
- Added data table simple link to ds-molecules-list.vue

### 🥼 Testing

- Tested the changes on http://localhost:8500/molecules/data-table-simple

- Whenever I add `table-responsive` class in es-data-table-simple the width does not exactly match the the width with [es-ds 2.0](https://design.energysage.dev/molecules/es-data-table-simple) designs. Whenever I dont have `table-responsive` class added it very closely matches the old designs so as of now I have not added the class to match the design. Does that work?
### 📝 Checklist


es-ds-components/components/es-data-table-simple.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <div :class="tableClass">
```
nathanielwarner, line 2: It seems like we should apply the `table-responsive` class here based on the [bootstrap docs](https://getbootstrap.com/docs/4.6/content/tables/#responsive-tables). I don't see any issue from that- were you seeing a problem with it put there?

es-ds-components/components/es-data-table-simple.vue
```diff
@@ -0,0 +1,61 @@
...
+
+const isSingleCol = computed(() => !props.fields);
+const computedItems = computed(() => {
+  if (!props.items.length && !useSlots().default) {
+    throw new Error('Neither items prop nor slot content is defined.');
+  }
+  // If there are no fields, just return the items as a single-column array
+  if (!props.fields) {
+    return props.items;
+  }
```
nathanielwarner, line 57: I notice in the old code there's these lines here:
```
if (this.fields.length !== this.items.length) {
    throw new Error('The number of fields and items must be the same.');
}
```
I'm wondering why you removed that check?

nathanielwarner (commented):
Looks pretty good! Just a few comments inline

es-ds-components/components/es-data-table-simple.vue
jaissica: When I was applying `table-responsive` where we have the `table-borderless` class the width of each cell was affected, so I have now moved `table-responsive` to the the `div` tag and it works fine now. `table-borderless` works fine when applied to `div` tag or `table` class so I followed the [Bootstrap docs](https://getbootstrap.com/docs/4.6/content/tables/#borderless-table) and added it to `table` class.

nathanielwarner: Nice!

nathanielwarner (approved):
LGTM! 🚀

PR #1480, opened 2024-08-19
CED-1745: EsFormMsg

nathanielwarner:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1745

### ❓ Type of change

### 📚 Description

Implements EsFormMsg. I first tried implementing it with PrimeVue Message, but found that the logic is simple enough that it's easier to just implement it from scratch, so that the component can be in the structure expected by our styling.

Currently, normal HTML buttons are used in the docs, since the EsButton is not available yet.

### 🥼 Testing

- Manual testing on http://localhost:8500/molecules/form-msg
- Verified functionality with VoiceOver

- General

### 📝 Checklist


es-ds-components/components/es-form-msg.vue
```diff
@@ -0,0 +1,69 @@
...
+    },
+});
+
+const emit = defineEmits(['hidden']);
+
+const hide = () => {
+    emit('hidden');
+}
+
+watch(() => props.show, (newValue, oldValue) => {
```
hroth1994, line 26: Other instances of `watch` in the code don't include `() =>` so I don't think it's necessary

nathanielwarner: What happens if you remove it? 🙂

es-ds-docs/pages/molecules/form-msg.vue
```diff
@@ -0,0 +1,184 @@
...
+const showHeaderSuccess = ref(false);
+const showSimpleError = ref(false);
+const showSimpleInfo = ref(false);
+const showSimpleSuccess = ref(false);
+
+const propTableRows = [
+    [
+        'show',
+        'false',
+        '(boolean) Whether to show the form message. This component will respond to changes of this prop, ' +
```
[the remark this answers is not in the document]
nathanielwarner: I don't think it matters a huge amount in general, but I think it's valuable in this instance because it's otherwise not clear whether you're supposed to pass the string "false"/"true" or a boolean

es-ds-components/components/es-form-msg.vue
```diff
@@ -0,0 +1,69 @@
...
+    if (newValue === true && oldValue === false) {
```
hroth1994, line 27: Could simplify this to `if (newValue && !oldValue) {`

nathanielwarner: They actually don't quite do the same thing, but I agree!

es-ds-components/components/es-form-msg.vue
hroth1994: Ah, I see the warnings now. Would we rather do it with `() =>` or define a ref to the prop to be used for `watch`? That's how it's done in radio button, so we can update that one or this one for consistency.

es-ds-docs/pages/molecules/form-msg.vue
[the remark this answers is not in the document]
hroth1994: In the docs for `Card` it has String variables surrounded by `''`. We could do that, or update the `Card` docs, I'm also noticing it doesn't use `ds-prop-table`

es-ds-components/components/es-form-msg.vue
nathanielwarner: Defining a ref to the prop doesn't work in and of itself, because then the ref is not updated when the prop is updated. Looking at es-radio-button, notice that they also define a watcher on a function that returns the prop, but they use that watcher to update the aforementioned detached ref. I think that's a more roundabout way to do it, and in any case I don't see that it matters in the context of this PR.

es-ds-docs/pages/molecules/form-msg.vue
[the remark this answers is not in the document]
nathanielwarner: Yeah I can see value in standardizing all the props tables to include types, or else doing that and having strings always have quotes around them.. at the moment it's kind of a free-for-all

es-ds-components/components/es-form-msg.vue
hroth1994: I think it matters for consistency - I'll make a note to update how it's done in `es-radio-button`

es-ds-docs/pages/molecules/form-msg.vue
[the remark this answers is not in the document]
hroth1994: Exactly 😅 I figure it'll be easier to define a standard now than after all of the components are complete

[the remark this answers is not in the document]
nathanielwarner: Hmmm yeah you've convinced me, I'll make that update now!

es-ds-components/components/es-form-msg.vue
nathanielwarner: I feel like my initial response was unnecessarily sassy 😭 What you didn't realize is how big a percentage of time spent on this code was spent on that line specifically, with me scratching my head over why I couldn't just watch a prop, and what the easiest workaround would be.

hroth1994 (approved):
LGTM! Thanks for adding the types to the other props tables 🌟

es-ds-components/components/es-form-msg.vue
hroth1994: It was a good learning exercise, and I don't mind a little sass! 💁‍♀️

PR #1481, opened 2024-08-19
feat: EsDataTable molecule for es-ds 3.0

angrocki:
### 🔗 Linked issue

[CED-1742: Create v3 EsDataTable using PrimeVue](https://energysage.atlassian.net/browse/CED-1742)

### ❓ Type of change

### 📚 Description

- Created data table molecule for es-ds 3.0
- Created es-data-table component
   - create a wrapper around PrimeVue DataTable
   - use pass through to fix style issues   
   - create functions and computed variables to use previous b-table props and features
      - Enable headers to be inferred from 1st item
      - Enable key to be made human readable (b-table feature) if label is not inputed
      - Enable sticky header, column and sorting
- Added DataTable link to ds-molecules-list.vue

### 🥼 Testing

- Tested locally at http://localhost:8500/molecules/data-table

- General implementation of styles and b-table features
- The data table may need more testing since it is used in `ArticleDataTable` which has over 300 results in Int
- Accessibility is not perfect
   - Chrome: reads sorted up/down on `Name` but does not read items below the scroll 
   - Safari: reads `Name` as clickable item but does not mention sorting and reads items below the scroll

### 📝 Checklist


nathanielwarner:
👀

es-ds-components/components/es-data-table.vue
```diff
@@ -0,0 +1,114 @@
+<template>
+    <DataTable
```
nathanielwarner, line 2: Nit: can we use kebab-case here and down below for Column?

es-ds-components/components/es-data-table.vue
```diff
@@ -0,0 +1,114 @@
...
+    return col?.label? col.label : convertKey(col?.key);
+};
+
+// Convert camel case and make first letter of string uppercase
+const convertKey = (string) => {
+    if (typeof string == 'string') {
+        const stringSpaced = string.replace(/([a-z])([A-Z])/g, '$1 $2');
+        return stringSpaced.charAt(0).toUpperCase() + stringSpaced.slice(1);
+    }
+    return string;
```
nathanielwarner, line 88: > Implemented uppercasing and converting camel case keys for the example table and ArticleDataTable [example](https://github.com/EnergySage/es-cms-articles/blob/4f94808e0fc1c33525597f7a075f685b24bdd5a8/components/storyblok/ArticleDataTable.vue#L42C16-L42C80), but b-table also converts kebab-case and snake_case to individual words

Is there any reason not to just copy BootstrapVue's logic for that into here?

es-ds-components/components/es-data-table.vue
```diff
@@ -0,0 +1,114 @@
...
+        if(typeof props.stickyHeader == 'string'){
+            return props.stickyHeader;
+        }
+        return '300px';
+    }
+    return '';
+});
+
+</script>
+<style>
```
nathanielwarner, line 109: Can we make sure to use `<style scoped lang="scss">` here? `.stickyFirstCol` probably needs to be wrapped in `:deep()` since it's being applied by the PrimeVue component.

es-ds-docs/pages/molecules/data-table.vue
```diff
@@ -0,0 +1,110 @@
...
+        <div class="mb-500">
+            <h2 class="mb-200">
+                Height limited to default 300px w/sticky header, sticky first column
+            </h2>
+            <es-data-table
+                sticky-header
+                sticky-first-col
+                :fields="dataTableFields"
+                :items="dataTableItems" />
+        </div>
```
nathanielwarner, line 39: I was about to comment that we should include component documentation, but then I realized that actually the demonstration is also the documentation in this case!

es-ds-docs/pages/molecules/data-table.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+    <div>
+        <h1>
+            Data table
+        </h1>
+        <p class="mb-500">
+            A responsive table using the <nuxt-link to="https://primevue.org/datatable/">PrimeVue DataTable</nuxt-link>.
```
nathanielwarner, line 7: Nit: these URLs should be https://v3.primevue.org/datatable/ since we're still on v3

es-ds-components/components/es-data-table.vue
```diff
@@ -0,0 +1,114 @@
...
+        :sortable="col?.sortable"
+        :frozen="checkFrozenRow(index)" alignFrozen="left"
+        :pt="{
+            sortIcon:{
+                style: 'display: none'
+            }, 
+            headerCell: (options)  => ({
+                class:[
+                    {
+                        'stickyFirstCol': options.props.frozen,
```
nathanielwarner, line 30: It seems like we also need to apply a class to make the cursor a pointer when hovering a sortable column

nathanielwarner: In order to tell the user it's clickable and maintain the current behavior

nathanielwarner (changes requested):
Good start on this!

es-ds-components/components/es-data-table.vue
angrocki: Changed to the code to use bootstraps logic!

es-ds-components/components/es-data-table.vue
angrocki: Good catch! I never noticed the pointers before!

es-ds-components/components/es-data-table.vue
angrocki: The deep comment saved me debugging time with the pointer!

nathanielwarner (approved):
LGTM! 🚀

PR #1488, opened 2024-08-22
feat: v3 EsModal

mpleroux:
### 🔗 Linked issue

[CED-1747: Create v3 EsModal using PrimeVue](https://energysage.atlassian.net/browse/CED-1747)

### ❓ Type of change

### 📚 Description

- New `es-modal` component using PrimeVue v3 `Dialog`
- Using HTML buttons for now until `EsButton` is ready

### 🥼 Testing

- Testing locally at http://localhost:8500/molecules/modal

- I used the Dialog's [emit callback](https://v3.primevue.org/dialog/#api.dialog.emits) `hide` instead of BootstrapVue's `@hidden` [event](https://bootstrap-vue.org/docs/components/modal#comp-ref-b-modal-events)
    - Does this seem right? If so, is breaking [downstream repos](https://github.com/search?q=org%3AEnergySage%20es-modal&type=code) that use `es-modal` unavoidable?
- I added the `size` prop, although the Dialog [prop](https://v3.primevue.org/dialog/#api.dialog.props) `breakpoints` could be more useful
- The `modal` prop doesn't dim the background by default. I'm looking into ways to implement this.

### 📝 Checklist


tomleo:
> The modal prop doesn't dim the background by default. I'm looking into ways to implement this.

Will that be in this PR or a follow-up one?

I also noticed there's a transition in the previous version, that's not present here. I'm not sure that's necessary but maybe worth double checking.

es-ds-docs/components/ds-molecules-list.vue
```diff
@@ -1,59 +1,40 @@
 <template>
```
tomleo, line 1: Looks like a lot of changes here, is this from a formatter?

tomleo (commented):
Overall looks good 👍, just open questions around transition & background dim to resolve I think

es-ds-docs/components/ds-molecules-list.vue
mpleroux: Yeah, I meant to add a note about that. I just added a list item for Modal and I haven't changed anything about my VSCode configuration in a while.

Is this list of [recommended extensions](https://github.com/EnergySage/es-ds/blob/main/.vscode/extensions.json) for es-ds still current? I noticed Vetur in there.

tomleo: Ahh yeah, I should create a ticket to update that. Vetur is for vue2, for the esds-3.0 branch we should be using https://marketplace.visualstudio.com/items?itemName=Vue.volar

mpleroux: I disabled `esbenp.prettier-vscode` it didn't try to reformat that file. Are we still using that one? It's in the list of extensions, but I thought I saw a comment from Ryan or someone else about it being obsolete.

nathanielwarner (changes requested):
Good start on this! There's a couple of inline comments, but overall the main outstanding thing is the background dim, which hasn't been implemented yet. Let me know if you want to strategize on that at some point.

es-ds-docs/components/ds-molecules-list.vue
mpleroux: After reviewing all my VSCode extensions and their settings I don't understand why Prettier keeps reformatting that file, unless that's the way it should be. I disabled the extension to add my changes for `Modal` and I won't touch it again.

mpleroux:
> > The modal prop doesn't dim the background by default. I'm looking into ways to implement this.
> 
> Will that be in this PR or a follow-up one?
> 
> I also noticed there's a transition in the previous version, that's not present here. I'm not sure that's necessary but maybe worth double checking.

It should be in this PR, but I don't know how to do it yet.

I've tried to get the transition working and so far haven't been successful.

mpleroux:
I believe I have addressed all the feedback items. I added a background mask using a similar color as the design system, and I used the transition classes and values from the PrimeVue component docs page.

tomleo (approved):
Looks good 👍 

I did noticed clicking outside of the modal does not dismiss it anymore, but that seems to be intentional by PrimeVue 🤷

es-ds-components/components/es-modal.vue
```diff
@@ -0,0 +1,107 @@
...
+import PrimeDialog from "primevue/dialog";
+
+const props = defineProps({
+    hideFooter: {
+        type: Boolean,
+        default: false,
+    },
+    size: {
+        type: String,
+        default: "lg",
```
nathanielwarner, line 11: Can we add the validator from the old one?
```
validator: (val) => ['sm', 'md', 'lg', 'xl'].includes(val),
```
I also notice that the default size has changed from `md` to `lg`- is this something we want?

es-ds-components/components/es-modal.vue
```diff
@@ -0,0 +1,107 @@
+<script setup lang="ts">
+import PrimeDialog from "primevue/dialog";
+
+const props = defineProps({
```
nathanielwarner, line 4: I notice that the `scrollable` prop is no longer here. I'm not sure what this one does in the old version- could you look into that and verify whether we should perhaps continue to support it?

(Side note: It's definitely fine to leave out `id` as an explicit prop since it's no longer required for PrimeVue and will be passed through if provided anyways)

es-ds-components/components/es-modal.vue
```diff
@@ -0,0 +1,107 @@
...
+@keyframes p-component-overlay-enter-animation {
+    from {
+        background-color: transparent;
+    }
+    to {
+        background-color: var(--maskbg);
+    }
+}
+
+@keyframes p-component-overlay-leave-animation {
```
nathanielwarner, line 99: It seems like the animation on leave is not working- is that something you've noticed? We have bigger fish to fry, but I figure it might be worth seeing if there's an easy fix

es-ds-components/components/es-modal.vue
```diff
@@ -0,0 +1,107 @@
...
+
+        <slot />
+
+        <template #footer v-if="!hideFooter">
+            <slot name="modal-footer" />
+        </template>
+    </prime-dialog>
+</template>
+
+<style lang="scss">
```
nathanielwarner, line 68: Could we use `scoped` here to avoid global CSS modifications? It'll be necessary to do `:deep()` on the selectors along with that to target the inner PrimeVue elements

es-ds-docs/pages/molecules/modal.vue
```diff
@@ -0,0 +1,99 @@
...
+
+<template>
+    <div>
+        <h1>Modal</h1>
+        <p>
+            Extended from
+            <nuxt-link href="https://v3.primevue.org/dialog/" target="_blank"> primevue modal </nuxt-link>
+        </p>
+        <div class="my-500">
+            <!-- TODO replace buttons with EsButton -->
```
nathanielwarner, line 29: Nit: can remove this comment

nathanielwarner (commented):
Looks generally good! There's a couple items which I think should be addressed, though none are critical. I verified that accessibility with VoiceOver is working properly as well.

es-ds-docs/pages/molecules/modal.vue
```diff
@@ -0,0 +1,99 @@
...
+                        varius.
+                    </p>
+                </template>
+                <template #modal-footer>
+                    <es-button @click="buttonsModalVisible = false">Cancel</es-button>
+                    <es-button class="ml-100" @click="buttonsModalVisible = false">Submit</es-button>
+                </template>
+            </es-modal>
+        </div>
+
```
nathanielwarner, line 91: Something I forgot- could we add the props documentation table here?

mpleroux:
I addressed all your feedback except for the mask and transition. I reverted my previous attempt to implement them because copying styles from the cascade and hardcoding class names seemed like the wrong approach.

For masks, I added a `background-color` style that seems to do what I want, although I don't know why. I can't get to work with a class in a scoped `style` element.

The modal in Design System v2 uses `background-color: rgba(34,38,51,.45)`, but that appears much darker when I try it in v3. My best guess for a matching mask color is `background-color: rgba(0, 0, 0, 0.075)`.

For transitions, nothing I have tried has worked. I added a few TODO notes in the code.

In the interest of moving on to a different PrimeVue component I'm going to merge this PR now. We can take another crack at implementing a transition another time.

es-ds-components/components/es-modal.vue
mpleroux: I ended up removing this code because it seemed like the wrong approach.

es-ds-components/components/es-modal.vue
mpleroux: I ended up removing this code because it seemed like the wrong approach.

PR #1491, opened 2024-08-23
feat: EsPopover molecule for es-ds 3.0 

jaissica:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-1749

### ❓ Type of change

### 📚 Description

 - Created popover molecule for es-ds 3.0
- Created es-popover component based on es-ds 2.0 - used pass through to fix style issues
- Added popover link to ds-molecules-list.vue

### 🥼 Testing

- Tested the changes on http://localhost:8500/molecules/popover

- The shape of the popover in primevue is different than the [bootstrap-vue popover](https://bootstrap-vue.org/docs/components/popover), does that work?
- Like in [bootstrap-vue popover](https://bootstrap-vue.org/docs/components/popover) I didn't find placement prop in primevue so I used css styling - `overlayPanelStyle` to achieve the same behavior.
Related issues - https://github.com/primefaces/primeng/issues/448, https://github.com/primefaces/primevue/issues/5058

### 📝 Checklist


es-ds-components/components/es-popover.vue
```diff
@@ -0,0 +1,157 @@
+
+<template>
+    <OverlayPanel ref="op" :dismissable="true" appendTo="body" :pt="popoverPt" :style="overlayPanelStyle">
```
tomleo, line 3: Although we don't have linting in place yet, I think we'll likely want to stick with kebab-case for component naming. I'd suggest switching this to `overlay-panel`

es-ds-components/components/es-popover.vue
```diff
@@ -0,0 +1,157 @@
...
+         class: `popover ${props.variant === 'light' ? 'es-popover-light':'es-popover-dark'}`,
+    },
+    content: {
+        class: 'es-popover-close',
+    },
+};
+</script>
+
+
+<style lang="scss">
```
tomleo, line 131: It looks like the previous design included `hover: pointer`, we might want to add that here

tomleo (commented):
It looks like we previously had an arrow connected to the popover, we might want to double check if that's still wanted.

nathanielwarner:
In making the arrows and shape work properly, I dove quite deep into the styling, and ended up doing a big cleanup to get rid of useless styles that overlap with each other.

I wasn't able to make the `placement` prop work either, as PrimeVue does not support this functionality. It seems like it's hardly used in our repos thankfully. I've removed this prop from the component and documentation.

nathanielwarner:
This should be all set for review now! I ended up moving the popover's arrow as far to the left as possible to help avoid the issue with things going offscreen on small displays with big popovers. The edge of the popover will still go a bit off screen in this case, but none of the content is affected.

nathanielwarner:
Noting as well that accessibility does not work with the popover, which is an issue [we identified with PrimeVue during the POC stage](https://energysage.atlassian.net/wiki/spaces/FG/pages/797016067/PrimeVue+Unstyled+Mode+POC#Accessibility).

tomleo:
👀

es-ds-components/components/es-popover.vue
```diff
@@ -0,0 +1,105 @@
...
+            targetElement.addEventListener(trigger, showPanel);
+        }
+    }
+});
+
+watch(
+    () => props.show,
+    (newVal) => {
+        if (newVal) {
+            const targetElement = document.getElementById(props.target);
```
tomleo, line 51: 🤔 should this be wrapped in an `if(import.meta.client)`?

tomleo (approved):
👍

es-ds-components/components/es-popover.vue
nathanielwarner: The onMounted hook is [not called during server side rendering](https://vuejs.org/api/composition-api-lifecycle#onmounted), so I don't think it matters

nathanielwarner: > This hook is typically used for performing side effects that need access to the component's rendered DOM, or for limiting DOM-related code to the client in a [server-rendered application](https://vuejs.org/guide/scaling-up/ssr).

PR #1493, opened 2024-08-23
CED-1726 exploration

tomleo:
Following https://nuxt.com/blog/eslint-module#nuxt-eslint-monorepo and https://github.com/antfu/eslint-config it looks like the `eslint-config` by [Anthony Fu](https://github.com/antfu) could be a low-lift way to get solid linting in our project.

Worth at least a look at how it formats everything.

You can see the changes it would apply here: https://github.com/EnergySage/es-ds/compare/esds-3.0-vue3-primevue...CED-1726-explore-formatter-applied?expand=1

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,0 +1,16 @@
+import antfu from '@antfu/eslint-config'
+
+export default antfu({
+    type: 'lib',
+    stylistic: {
+        indent: 4,
+        quotes: 'single',
+    },
+    typescript: {
+        tsconfigPath: './es-ds-docs/tsconfig.json',
+    },
+    vue: true,
+    formatters: {
+        html: true, // uses Prettier
+    },
+})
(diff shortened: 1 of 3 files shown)


tomleo:
> Why use this `@antfu` library versus just doing it ourselves? https://eslint.org/docs/latest/use/configure/configuration-files

Don't really have a strong opinion, just seemed like something worth looking at, we can go either way. I think one advantage of using something like this is the number of contributors and first-class support for nuxt.

nathanielwarner:
@tomleo I do agree it'd be nice not to have to handle things like the Nuxt implicit globals ourselves, and it's nice how little configuration this one has. There are some things that this configuration does differently from our existing one, such as not having semicolons after statements, and using PascalCase instead of kebab-case for components. Presumably that's easy to adjust, and of course our existing style isn't set in stone anyways.

That said, Raf already has a PR up which has also handled most of the linting errors that required manual intervention. I think it's a step in the right direction, and might be easier to start from there rather than starting over, particularly when it comes to those manual interventions like fixing unused variables throughout the project. But perhaps using antfu's library could be integrated into Raf's PR as well, along with a small amount of changes

PR #1499, opened 2024-08-28
fix: vite version

tomleo:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1830

### ❓ Type of change

### 📚 Description

There seems to be some instability around vite optimizations for certain versions of vite & nuxt. This PR sets an explicit vite version and adds an explicit `optimizeDeps` argument to avoid slow optimization loops during local-dev.

### 🥼 Testing

- Local Dev

- Has anyone else run into this issue yet?

### 📝 Checklist


es-ds-docs/nuxt.config.ts
```diff
@@ -10,4 +10,27 @@ export default defineNuxtConfig({
 
     // https://nuxt.com/docs/getting-started/deployment#static-hosting
     ssr: true,
+
+    // Per suggestion from https://github.com/nuxt/nuxt/issues/26783#issuecomment-2058966800
+    // See also:
+    // - https://github.com/nuxt/nuxt/issues/24196#issuecomment-1853870504
+    vite: {
+        optimizeDeps: {
+            include: [
```
tomleo, line 19: This list would eventually include all primevue components we're wrapping, and could be moved into the `es-ds-components` later.

nathanielwarner (commented):
This does seem to fix that annoying issue where dependencies have to get optimized during load.. perhaps `@vuelidate/core, @vuelidate/validators` should be added as well since I see those. But also, maybe we should be updating Nuxt instead? The latest version of Nuxt has a dependency on the new version of Vite you're specifying, so we get that "for free" and avoid any incompatibilities between the old version of Nuxt and the new version of Vite.

nathanielwarner (approved):
Awesome! 👍 Loving the increased performance

PR #1502, opened 2024-08-29
feat: updated default layout to match normal max width; created fullWidthComponent layout for edge cases

ericdouglaspratt:
### 🔗 Linked issue

- n/a

### ❓ Type of change

### 📚 Description

- In order to better be a showcase and example for all EnergySage teams on how our layout and max-width container works, most pages on the design system site should adhere to our own max width container conventions
- When I originally added the side nav at larger breakpoints, I deliberately broke it out of the container so that the organisms that are intended to render at full width (e.g. CTA banner, support card) are able to be previewed in a manner like how they're supposed to appear on a page - however there are a limited number of these edge cases such that this behavior should not be the default layout - I've restored the default layout to a normal max width container and preserved the container breakout of the side nav in the fullWidthComponent layout for use when we build those organism pages
- Also fixed top padding on mobile

### Before
Screen Shot 2024-08-29 at 3 04 47 PM

### After
Screen Shot 2024-08-29 at 3 04 59 PM

### 🥼 Testing

- Tested locally

- Overall

### 📝 Checklist

### DIFF
--- a/es-ds-docs/layouts/default.vue
+++ b/es-ds-docs/layouts/default.vue
@@ -1,24 +1,22 @@
 <template>
-    <div>
-        <div class="d-flex justify-content-center">
-            <div class="ds-side-nav d-none d-xl-block flex-shrink-0 p-100">
-                <ds-link-list />
-            </div>
-            <b-container class="pt-xl-100 mx-0">
-                <b-row class="mb-100">
-                    <b-col cols="12">
+    <div class="d-flex justify-content-center">
+        <b-container class="pt-100 mx-0">
+            <b-row>
+                <b-col class="d-none d-xl-block" xl="3">
+                    <div class="ds-side-nav flex-shrink-0">
+                        <ds-link-list />
+                    </div>
+                </b-col>
+                <b-col xl="9">
+                    <div class="mb-100">
                         <es-breadcrumbs :items="breadcrumbs" />
-                    </b-col>
-                </b-row>
-                <b-row>
-                    <b-col
-                        class="mb-300"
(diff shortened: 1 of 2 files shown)


hroth1994:
👀

es-ds-docs/layouts/default.vue
```diff
@@ -1,24 +1,22 @@
...
-    <div>
-        <div class="d-flex justify-content-center">
-            <div class="ds-side-nav d-none d-xl-block flex-shrink-0 p-100">
-                <ds-link-list />
-            </div>
-            <b-container class="pt-xl-100 mx-0">
-                <b-row class="mb-100">
-                    <b-col cols="12">
+    <div class="d-flex justify-content-center">
+        <b-container class="pt-100 mx-0">
```
hroth1994, line 3: The `16px` margin was coming from the top nav before, so just making a note we'll need to remove `mb-100` from that element when it gets added back.

hroth1994 (approved):
left 1 comment but LGTM! 🌟

es-ds-docs/layouts/default.vue
ericdouglaspratt: Yeah we were using the default styling of Bootstrap Vue's navbar component before, which I'm not sure we should put effort into bringing forward into ESDS 3.0, as we have no Figma designs for it. Breadcrumbs and category pages may be enough to navigate on mobile, but once we put more thought into navs in general (and potentially rebuild the EsNavBar to be accessible), we may discover some better PrimeVue tools we can use for this.

PR #1504, opened 2024-08-29
CED-1748: esPagination primeVue component

lgeggleston:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1748

### ❓ Type of change

### 📚 Description

- Pagination component based on [primeVue paginator](https://v3.primevue.org/paginator)
- corresponding documentation page
- Updated styles using Passthrough, which could be simplified because there were less layers than Bootstrap pagination, and also removed some styles that would no longer be used (i.e. for page # buttons since we are replacing with "X of Y" display)

-

### 📝 Checklist


nathanielwarner:
👀

es-ds-components/components/es-pagination.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+    <Paginator
```
nathanielwarner, line 2: Nit: We don't yet have linting set up, but can we switch this to kebab-case like the rest of the components? i.e. `<paginator>`

es-ds-components/components/es-pagination.vue
```diff
@@ -0,0 +1,46 @@
...
+        v-bind="$attrs"
```
nathanielwarner, line 3: Another nit: v-bind attrs is unnecessary [because there's only one root element](https://vuejs.org/guide/components/attrs#attribute-inheritance)

es-ds-docs/pages/molecules/pagination.vue
```diff
@@ -0,0 +1,56 @@
...
+            <ul
+                id="paginated-list">
+                <li
+                    v-for="item in items.slice(first, first + perPage)"
+                    :key="item">
+                    {{ item }}
+                </li>
+            </ul>
+            <es-pagination
+                v-model:first="first"
```
nathanielwarner, line 22: So currently people just pass a v-model with the current page, rather than the "first" which corresponds to the record number that's at the start of the current page. How would you feel about matching the existing API we have for pagination? Our best practices are still being developed, but the [current guidance](https://energysage.atlassian.net/wiki/spaces/CED/pages/1009025072/PrimeVue+and+Nuxt+3+implementation+guide#General-guidance) is "to provide a wrapper component of the same name in ESDS 3.0 that matches the props API of the original component... The only exception to this is if creating an ESDS-2.0-style wrapper component around a PrimeVue component would be so different from how the PrimeVue component works as to take a long time or be overly complex."

I'm happy to strategize with you on this- I think it would just be a matter of defining a model on `es-pagination` which would then be translated to and from PrimeVue's format based on the number of items per page.

es-ds-docs/pages/molecules/pagination.vue
```diff
@@ -0,0 +1,56 @@
...
+                :total-rows="rows"
+                :per-page="perPage"
+                list-id="paginated-list"
+                align="center" />
+        </div>
+
```
nathanielwarner, line 28: Could we provide the props documentation here? That's especially important if we do change the API in any way.

nathanielwarner (commented):
Nice work on this! I think the styling and layout is a positive change. The main thing I'm not sure about is the API, though after doing a search through GitHub I see that actually there's only 3 downstream usages of `<es-pagination>`, so changing the API might be fine if it takes a non-trivial amount of time to match the existing one. I do think we should at least document how to use it though.

nathanielwarner (commented):
Oh and one more thing- can we include `primevue/paginator` in nuxt.config.ts, in the vite -> optimizeDeps -> include, so that we don't get that annoying `✨ new dependencies optimized: primevue/paginator` that reloads the page repeatedly when spinning up the dev server?

es-ds-components/components/es-pagination.vue
lgeggleston: ah good point

es-ds-docs/pages/molecules/pagination.vue
lgeggleston: That's fair! I was originally thinking it should be a really easy switch downstream, but it's also not nothing and would require the arithmetic to change for how list items are displayed.  Implemented by defining a model for `currentPage` from the parent and setting up reactivity as you suggested

es-ds-docs/pages/molecules/pagination.vue
lgeggleston: Good idea! added

lgeggleston:
Thanks for suggestions! Updated to match the current API in terms of the `v-model` input, so it shouldn't require changes downstream. 
@nathanielwarner one more question - I tested that I can navigate and click the prev/next buttons using keyboard, but  is there anything else you would suggest checking accessibility-wise?

es-ds-components/components/es-pagination.vue
```diff
@@ -0,0 +1,59 @@
...
+        :aria-controls="listId"
+        :pt="passThrough">
+        <template #prevpagelinkicon><IconChevronLeft /></template>
+        <template #nextpagelinkicon><IconChevronRight /></template>
+    </paginator>
+</template>
+
+<script setup lang="ts">
+import Paginator from 'primevue/paginator';
+import { ref, watch } from 'vue';
```
nathanielwarner, line 18: Nit: We don't need to import `ref` and `watch`

PR #1512, opened 2024-09-04
Ced 1746 es carousel

hroth1994:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1746

### ❓ Type of change

### 📚 Description

Putting a PR up to get initial feedback on the new EsCarousel component that uses PrimeVue's Carousel. I'm still working on getting a few things working (see Feedback section).

### 🥼 Testing

Tested locally on http://localhost:8500/molecules/carousel - rotated through the carousel on different viewports and pausing autoplay

A few things I'm still trying to get working:
- Getting the carousel to append the first element at the end rather than restarting the carousel
- When stopping autoplay, stay in the same position rather than restarting the carousel
- Refreshing shows a different order of items

### 📝 Checklist

### DIFF
--- a/es-ds-components/components/es-carousel.vue
+++ b/es-ds-components/components/es-carousel.vue
@@ -1,0 +1,61 @@
+
+<script setup lang="ts">
+import Carousel from 'primevue/carousel';
+
+const responsiveOptions = ref([
+    {
+        breakpoint: '1400px',
+        numVisible: 2,
+        numScroll: 1
+    },
+    {
+        breakpoint: '1199px',
+        numVisible: 3,
+        numScroll: 1
+    },
+    {
+        breakpoint: '767px',
+        numVisible: 2,
+        numScroll: 1
+    },
+    {
+        breakpoint: '575px',
+        numVisible: 1,
+        numScroll: 1
+    }
+]);
+
+
+defineProps({
+    items: {
+        type: Array,
+        default: [],
+        required: true,
+    },
+});
+
+</script>
+
+<template>
+    <Carousel :value="items" :numVisible="3" circular :pt="{
+            itemsContent: {
+                style: 'overflow: hidden;',
+            },
+            itemsContainer: {
+                class: 'd-flex',
+            },
+            item: {
+                class: 'm-1',
+            }
+        }">
+        <template #item="slotProps">
+            <es-card :style="{ width: '400px' }" class="h-100 flex align-items-center justify-content-center">
+                {{ slotProps.data }}
+            </es-card>
+        </template>
+    </Carousel>
+</template>
+
+<style lang="scss" scoped>
+
+</style>
--- a/es-ds-components/components/es-carousel.vue
+++ b/es-ds-components/components/es-carousel.vue
@@ -3,30 +3,48 @@
 import Carousel from 'primevue/carousel';
 
 const responsiveOptions = ref([
+// (xs: 0, sm: 576px, md: 768px, lg: 992px, xl: 1200px)
     {
-        breakpoint: '1400px',
-        numVisible: 2,
+        breakpoint: '1199px',
+        numVisible: 5,
         numScroll: 1
     },
     {
-        breakpoint: '1199px',
-        numVisible: 3,
+        breakpoint: '991px',
+        numVisible: 4,
         numScroll: 1
     },
     {
         breakpoint: '767px',
-        numVisible: 2,
+        numVisible: 3,
         numScroll: 1
     },
     {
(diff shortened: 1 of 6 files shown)


tomleo:
👀

es-ds-components/components/es-carousel.vue
```diff
@@ -0,0 +1,160 @@
...
+        default: () => [],
+        required: true,
+    },
+});
+
+const autoplayInterval = ref(props.autoscroll ? 3000 : 0);
+const key = ref('');
+
+const stopAutoplay = () => {
+    if (props.autoscroll) {
```
tomleo, line 51: Could this be `if(autoplayInterval.value > 0)` to ensure it will always `stopAutoPlay`?

es-ds-components/components/es-carousel.vue
```diff
@@ -0,0 +1,160 @@
...
+        autoplayInterval.value = 0;
+        key.value = 'stopAutoplay';
+    }
+};
+
+onMounted(() => {
+    document.addEventListener('keyup', (e) => {
```
tomleo, line 58: Without event delegation, could this potentially capture other `keyup` events? Could we use a vue [@keyup.enter=](https://vuejs.org/guide/essentials/event-handling.html#key-modifiers)?

es-ds-docs/pages/molecules/carousel.vue
```diff
@@ -0,0 +1,209 @@
+<script setup lang="ts">
+const items = ref([
+    {
+        image_url:
+            'https://edge.disstg.commercecloud.salesforce.com/dw/image/v2/BDDP_STG/on/demandware.static/-/Sites-masterCatalog/default/Products/I-PUPEVCHAR-NEMA.jpg',
+        model_name: 'Wallbox Pulsar Plus 40A',
+        price: '$649.00',
+        productUrl: 'https://equipment.energysage.com/ev-chargers/PUP1-U-1-5-N-002-A',
```
tomleo, line 8: Could we use fake data here for the example? We could use something like https://placekitten.com/ to generate images in the sizes we need.

tomleo (commented):
Looking good! A few comments inline.

Also I think a Carousel might be better [categorized as an Organism](https://atomicdesign.bradfrost.com/chapter-2/#organisms) given it's composed of other molecules

es-ds-components/components/es-carousel.vue
nathanielwarner: > Could we use a vue [@keyup.enter=](https://vuejs.org/guide/essentials/event-handling.html#key-modifiers)?

Unfortunately no, because it needs to capture the escape key pressed while focus is anywhere on the page

nathanielwarner (commented):
My one suggestion at this point is to add `primevue/carousel` to the `optimizeDeps` in the nuxt config in `es-ds-components`, to stop the annoying auto-refresh while loading

nathanielwarner (commented):
I like the cats! Other than merge conflicts, I think the issues you identified are the only ones I see. Do you think it's gonna be possible to address them with the PrimeVue carousel? I guess it's a product question as to whether it's acceptable, but they don't affect our current carousels.

es-ds-docs/nuxt.config.ts
```diff
@@ -8,5 +8,5 @@ export default defineNuxtConfig({
     // https://nuxt.com/docs/getting-started/deployment#static-hosting
     ssr: true,
 
-    modules: ['./modules/auto-import-eslint.ts'],
+    modules: ['./modules/auto-import-eslint.ts', '@nuxt/image'],
```
tomleo, line 11: Could `@nuxt/image` live in the `es-ds-components` layer?

tomleo (approved):
Looks good, only question is where `nuxt/image` should live (in the layer or in the docs)

es-ds-docs/nuxt.config.ts
ericdouglaspratt: Maybe, but it is possible to use design system components in a micro frontend without needing nuxt image, so maybe we'd better leave that up to downstream repos. I'd guess _most_ downstream repos would end up using nuxt image, but maybe not all.

hroth1994:
> I like the cats! Other than merge conflicts, I think the issues you identified are the only ones I see. Do you think it's gonna be possible to address them with the PrimeVue carousel? I guess it's a product question as to whether it's acceptable, but they don't affect our current carousels.

I think it could be possible with more investigation, and I'm also curious to see if it's improved in the next version.

PR #1518, opened 2024-09-10
feat: v3 slider

mpleroux:
### 🔗 Linked issue

[CED-1755: Create v3 EsSlider using PrimeVue](https://energysage.atlassian.net/browse/CED-1755)

### ❓ Type of change

### 📚 Description

- New Slider component extended from [PrimeVue Slider](https://v3.primevue.org/slider/)
- Recreated styling from v2 Slider using pass through styles
- Created a [fake tooltip](https://github.com/primefaces/primevue/issues/1895#issuecomment-1573486729) using `::before` pseudo-element

### 🥼 Testing

- Tested locally at http://localhost:8500/molecules/slider

### 📝 Checklist


es-ds-components/components/es-slider.vue
```diff
@@ -0,0 +1,107 @@
...
+});
+
+// allow use of v-model on this component
+const model = defineModel<number>();
+</script>
+
+<template>
+    <div class="mb-3">
+        <slider
+            v-model="model"
```
hroth1994, line 36: `min`, `max`, and `step` need to be included as props

es-ds-components/components/es-slider.vue
```diff
@@ -0,0 +1,107 @@
+<script setup lang="ts">
+import Slider from 'primevue/slider';
+
+defineProps({
+    min: {
+        type: Number,
+        required: true,
+        default: 0,
+    },
+    max: {
```
hroth1994, line 10: I don't think it's a good idea to change the props that this component is expecting.
https://github.com/search?q=org%3AEnergySage%20esslider&type=code

es-ds-docs/pages/molecules/slider.vue
```diff
@@ -0,0 +1,63 @@
...
+    const docSource = await import('./slider.vue?raw');
+    /* eslint-enable import/no-webpack-loader-syntax, import/no-self-import */
+
+    compCode.value = $prism.normalizeCode(compSource.default);
+    docCode.value = $prism.normalizeCode(docSource.default);
+    $prism.highlight();
+}
+
+const propTableRows = [
+    ['min', 'Number', 'true', 'Minimum value of slider.'],
```
hroth1994, line 23: Props table default values need to be updated

es-ds-docs/pages/molecules/slider.vue
```diff
@@ -0,0 +1,63 @@
...
+        <p>
+            Extended from
+            <nuxt-link
+                href="https://v3.primevue.org/slider/"
+                target="_blank">
+                PrimeVue Slider
+            </nuxt-link>
+        </p>
+
+        <es-slider
```
hroth1994, line 43: Put this slider within the div below it for better spacing

hroth1994 (commented):
Made some initial comments - I like how the tooltip is looking so far!

es-ds-docs/pages/molecules/slider.vue
```diff
@@ -0,0 +1,66 @@
...
+const min = ref(0);
+const max = ref(1000);
+const step = ref(50);
+const ariaLabel = ref('Sample aria-label');
+
+const { $prism } = useNuxtApp();
+const compCode = ref('');
+const docCode = ref('');
+
+if ($prism) {
```
nathanielwarner, line 12: Can we apply the prism fix here that was applied to other components? (wrapping the if block in an async function inside onMounted)

mpleroux:
I've addressed all the feedback mentioned here except for changing the props.

I'm not sure what to do about old [vue-slider-component](https://nightcatsama.github.io/vue-slider-component/#/api/props) props that aren't supported by the [PrimeVue slider](https://v3.primevue.org/slider/) component.

These don't have equivalents in PrimeVue:

- `startingValue` (starting value of slider thumb)
- `data` (array that contains valid steps)
- `marks` (function that chooses where to add marks to the track)

This one would be useful:

- `tooltipFormatter` (function that modifies tooltip value) 

But I'm not sure if it's possible to modify the `aria-valuenow` I'm referencing in the `style` tag.

And it would be good to support this event somehow:

- `@change` callback (using [@SliderSlideEndEvent](https://v3.primevue.org/slider/#api.slider.emits.slideend)?)

es-ds-components/components/es-slider.vue
```diff
@@ -0,0 +1,139 @@
...
+
+// "tooltip"
+:deep(.slider-handle::before) {
+    align-items: center;
+    background-color: variables.$blue-600;
+    border-radius: 50%;
+    bottom: 27px;
+    box-shadow: 0 1px 6px 0 rgba(34, 38, 51, 0.25);
+    color: variables.$white;
+    content: attr(aria-valuenow);
```
nathanielwarner, line 129: This is a really clever CSS solution to show the value on the tooltip! I think we actually can iterate on this using Vue 3's support for [v-bind in CSS](https://vuejs.org/api/sfc-css-features.html#v-bind-in-css). By controlling the value from the component, this would allow us to support the `tooltip-formatter` prop.

es-ds-docs/pages/molecules/slider.vue
```diff
@@ -0,0 +1,70 @@
...
+                :min="min"
+                :max="max"
+                :step="step"
+                :starting-value="150"
+                :aria-label="ariaLabel"
+                :label-formatter="(val) => `$${val}`" />
+        </div>
+
+        <div class="mb-500">
+            <h2>Slider props</h2>
```
nathanielwarner, line 60: Nit: es-slider props

nathanielwarner (changes requested):
This is a really good start! I think with a few tweaks we can cover pretty much all existing use cases.
- The CSS adjustment I mentioned inline to support `tooltip-formatter`
- In our codebase, `marks` is only used to add marks to the start and end of the slider (with [one exception in CDGM](https://github.com/EnergySage/es-cdgm/blob/0a9060f1c228a4e3d19edbce47e7d4fe9d5b84ad/frontend/pages/developer-admin/projects/_slug.vue#L444)). Do you think it would be relatively straightforward to allow this specific use case, perhaps with props of `startLabel` and `endLabel`? We'd want to leave a note about migration.
- It doesn't seem like there's a straightforward way to support the `data` prop that allows a non-constant step in the slider.. Except if we do some trickery with `v-model` 🤔. I wonder if it'd be possible to not pass `step` to PrimeVue slider at all, but instead dynamically set the model value to the nearest allowed value after the user lets go of the slider. I'll also note that we actually don't seem to use `step` at all in our repos, so maybe we should just get rid of this prop.
  - One other benefit to dynamically setting the model value after the user lets go is to provide a smoother experience.. right now, PrimeVue's default behavior feels kind of choppy because the slider doesn't move at all until you reach the next step.

For reference, the existing use cases:
- [kebab-case](https://github.com/search?q=org%3AEnergySage%20%3Ces-slider&type=code)
- [PascalCase](https://github.com/search?q=org%3AEnergySage+%3CEsSlider&type=code)

es-ds-docs/pages/molecules/slider.vue
```diff
@@ -0,0 +1,70 @@
...
+
+const { $prism } = useNuxtApp();
+const compCode = ref('');
+const docCode = ref('');
+
+onMounted(async () => {
+    if ($prism) {
+        /* eslint-disable import/no-webpack-loader-syntax, import/no-self-import */
+        const compSource = await import('@energysage/es-ds-components/components/es-slider.vue?raw');
+        const docSource = await import('./slider.vue?raw');
```
hroth1994, line 25: Just needs one comment above this line `// eslint-disable-next-line import/no-self-import`

mpleroux:
@nathanielwarner @hroth1994 I've put several more hours into this PR and I'm just stuck.

The [v-bind in CSS](https://vuejs.org/api/sfc-css-features.html#v-bind-in-css) technique you mentioned is straightforward, and I can get it to work with a regular CSS property like `color`, but it doesn't work with `content` no matter what I try.

A [similar technique using CSS variables](https://stackoverflow.com/questions/50625973/v-bindstyle-pseudo-element-after-content-icon) also doesn't work for `content`.

I don't understand what would be necessary to manipulate `v-model` in order to make the animation smoother.

At this point I'd like to merge this PR and create a new ticket for the followup tasks mentioned in these comments.

mpleroux:
ChatGPT has some suggestions for using `data-`attributes instead. I'm looking into that.

mpleroux:
I merged [Tom's PR](https://github.com/EnergySage/es-ds/pull/1548) to use `v-bind` with CSS `content` property. Thanks, Tom!

And I restored the `tooltipFormatter` to maintain the API.

nathanielwarner (approved):
Nice work on this! I think we've spent enough time on this for now, and what you have pretty much covers the existing use cases. Looking back at my note, I'm not so sure anymore about whether it's actually feasible to support a non-constant step the way the old one does...

Would you feel comfortable making a followup ticket? The main thing is to evaluate current usages of `<es-slider>` (including `<EsSlider>`) and see if we can close any gaps in functionality, and make a section in this documentation for any migration steps that will be necessary in the downstream repos.

PR #1519, opened 2024-09-10
feat: add EsFormTextarea v3

hroth1994:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1758

### ❓ Type of change

### 📚 Description

Adds molecule `EsFormTextarea` based off of [existing implementation](https://design.energysage.dev/molecules/es-form-textarea/) that uses HTML element `textarea`. I initially tried using [PrimeVue's](https://v3.primevue.org/textarea/) `Textarea` component but the unstyled mode is essentially the same as `textarea`.

### 🥼 Testing

Tested locally at http://localhost:8500/molecules/form-textarea ensuring the entered value propagates across all input fields, excluding the success state fields which have their own message

- A few things to note:
- I tried `v-bind="$attrs"` but ended up using `modelValue` when that worked with neither `Textarea` nor `textarea`
- hover, active, and focus pseudo-classes match existing implementation rather than ESDS 2.0 design
- border width matches design, not existing implementation (existing: .556px, ESDS 2.0 design: 1px)

### 📝 Checklist

### DIFF
--- a/es-ds-components/components/es-form-textarea.vue
+++ b/es-ds-components/components/es-form-textarea.vue
@@ -29,14 +29,6 @@ const props = defineProps({
         required: true,
     },
     /**
-     * modelValue
-     */
-    modelValue: {
-        type: String,
-        required: true,
-        default: '',
-    },
-    /**
      * state
      */
     state: {
@@ -44,20 +36,7 @@ const props = defineProps({
         default: null,
     },
 });
-
-const emit = defineEmits(['update:modelValue']);
-const localValue = ref(props.modelValue);
-// Watch for changes to the local value and emit them
-watch(localValue, (newValue) => {
-    emit('update:modelValue', newValue);
-});
-// Watch for prop changes and update the local value
-watch(
-    () => props.modelValue,
-    (newValue) => {
-        localValue.value = newValue;
-    },
-);
+const model = defineModel<string>();
 
 const slots = useSlots();
 
@@ -69,15 +48,15 @@ const hasError = () => !!slots.errorMessage;
 <template>
     <div
         class="input-wrapper justify-content-end"
-        :required="required">
+        :required="props.required">
         <!-- eslint-disable-next-line vuejs-accessibility/label-has-for -->
         <label
-            :for="id"
+            :for="props.id"
             class="label justify-content-start">
             <slot name="label" />
 
             <span
-                v-if="required"
+                v-if="props.required"
                 class="text-danger">
                 *
             </span>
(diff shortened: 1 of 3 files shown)


nathanielwarner:
👀

es-ds-components/components/es-form-textarea.vue
```diff
@@ -0,0 +1,145 @@
...
+        type: String,
+        required: true,
+    },
+    /**
+     * modelValue
+     */
+    modelValue: {
+        type: String,
+        required: true,
+        default: '',
```
nathanielwarner, line 37: This custom code for `modelValue` and `update:modelValue` and `localValue` is not necessary.. All that's needed is to use `defineModel`, and set that value as `v-model` on the native `textarea`.

es-ds-docs/pages/molecules/form-textarea.vue
```diff
@@ -0,0 +1,165 @@
+<script setup>
+const docTextarea = ref('');
+const successValue = 'My experience was great!';
```
nathanielwarner, line 3: This should be a `ref` since it's reactive state, to avoid errors

es-ds-docs/components/ds-molecules-list.vue
```diff
@@ -33,6 +33,9 @@
         <li>
             <ds-link to="/molecules/form-message"> Form message </ds-link>
         </li>
+        <li>
+            <ds-link to="/molecules/form-textarea"> Form textarea </ds-link>
```
nathanielwarner, line 37: Wondering why this is "Form textarea" instead of "Textarea" as the old one was?

es-ds-components/components/es-form-textarea.vue
nathanielwarner: To clarify, what I'm suggesting:
- remove `modelValue` prop
- Add `const model = defineModel<string>();`
- Remove `const emit...`
- Remove `const localValue...`
- Remove `watch(localValue...`
- Remove `watch(...`
- On the `textarea` component, replace `v-model="localValue"` with `v-model="model"`. Also apply `v-bind="$attrs"` as was stated on line 3

nathanielwarner (commented):
Looks pretty good! Just a cleanup and a fix.

es-ds-docs/components/ds-molecules-list.vue
hroth1994: Good question - I was matching the component name but could get behind just putting Textarea

es-ds-components/components/es-form-textarea.vue
```diff
@@ -0,0 +1,124 @@
+<script setup lang="ts">
+// Prevents attributes from being applied to first <div>
+// v-bind="$attr" is on the input instead
+defineOptions({
+    inheritAttrs: false,
+});
+
+const props = defineProps({
```
nathanielwarner, line 8: Nit: the props are only used in the template, so you don't need to save the variable. You can instead just call `defineProps({...` and reference the props without `props.` in the template

nathanielwarner (approved):
Looks good! Just the one nit.

es-ds-components/components/es-form-textarea.vue
hroth1994: I had that originally but after adding `defineModel` it made the props have this error:
```
Property 'required' does not exist on type 'CreateComponentPublicInstanceWithMixins<ToResolvedProps<__VLS_PublicProps, { "update:modelValue": (modelValue: string) => any; }>, { model: ModelRef<string | undefined, string, string | undefined, string | undefined>; hasSuccess: () => boolean; hasMessage: () => boolean; hasError: () => boolean; }, ... 23 more ..., ...'.ts-plugin(2339)
```

nathanielwarner: Hmm I'm not seeing that error. In any case not a big deal, don't wanna hold up merging this if it's hard to address

hroth1994: Removed! It must be an extension I have installed.

PR #1520, opened 2024-09-10
CED-1762 create v3 EsViewMore

lgeggleston:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1762

### ❓ Type of change

### 📚 Description

- Create `EsViewMore` in Vue 3, no need for primeVue component
- install dependency `html-truncate`
- Removed a piece of logic from original version of this component that seems to pass down click functionality from the parent if given and replace default show more/show less (`if (this.$listeners && this.$listeners.click) ...`).  ~However, a check through ES repo usage of this component showed this is not currently used, and I don't think it makes much sense to continue, so removed in this case.~.  
    - Update: see Feedback for reasoning
    - To note: after looking into migrating this bit to Vue3 it also seems either deprecated or messy to work around: https://github.com/vuejs/rfcs/discussions/397

- The click override thing is actually used for a real reason in one place - EsReview ([ to open a modal in CDGM](https://github.com/EnergySage/es-cdgm/blob/0a9060f1c228a4e3d19edbce47e7d4fe9d5b84ad/frontend/pages/projects/_slug/index.vue#L422)).  However, we decided to [defer recreating EsReview](https://energysage.atlassian.net/browse/CED-1791) since it's only used in CDGM for now.  I think it may actually be simpler to replace that use in `EsReview` with a truncated span+button, since that's the only part of the component really used - but we only need to do that if CDGM is getting upgraded to vue3 at all.
    - does it sounds alright if I just add this note to the Review ticket, in case gets done in the future?

### 📝 Checklist

### DIFF
--- a/es-ds-docs/package.json
+++ b/es-ds-docs/package.json
@@ -27,7 +27,8 @@
         "nuxt-primevue": "^3.0.0",
         "primevue": "^3.53.0",
         "prismjs": "^1.29.0",
-        "vue": "^3.4.34"
+        "vue": "^3.4.34",
+        "html-truncate": "^1.2.2"
     },
     "devDependencies": {
         "@eslint/js": "^9.9.0",
--- a/es-ds-docs/package.json
+++ b/es-ds-docs/package.json
@@ -18,6 +18,7 @@
     "dependencies": {
         "@energysage/es-ds-components": "^3.0.0-alpha.6",
         "@energysage/es-ds-styles": "^3.0.0-alpha.6",
+        "@nuxt/image": "^1.8.0",
         "@nuxt/kit": "^3.13.1",
         "@nuxtjs/google-fonts": "^3.2.0",
         "@vuelidate/core": "^2.0.3",
--- a/es-ds-docs/package.json
+++ b/es-ds-docs/package.json
@@ -28,8 +28,7 @@
         "nuxt-primevue": "^3.0.0",
         "primevue": "^3.53.0",
         "prismjs": "^1.29.0",
-        "vue": "^3.4.34",
-        "html-truncate": "^1.2.2"
+        "vue": "^3.4.34"
     },
     "devDependencies": {
         "@eslint/js": "^9.9.0",
--- a/es-ds-docs/pages/molecules/support.vue
+++ b/es-ds-docs/pages/molecules/support.vue
@@ -59,8 +59,8 @@ const supportProps = [
 
         <ds-doc-source
             :comp-code="compCode"
-            comp-source="es-ds-components/components/es-pagination.vue"
+            comp-source="es-ds-components/components/es-support.vue"
             :doc-code="docCode"
-            doc-source="es-ds-docs/pages/molecules/pagination.vue" />
+            doc-source="es-ds-docs/pages/molecules/support.vue" />
     </div>
 </template>
--- a/es-ds-docs/pages/molecules/view-more.vue
+++ b/es-ds-docs/pages/molecules/view-more.vue
@@ -51,11 +51,12 @@ const viewMoreProps = [
     <div>
         <h1>View more</h1>
         <p>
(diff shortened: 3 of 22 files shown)


hroth1994:
👀

es-ds-docs/package.json
```diff
@@ -28,7 +28,8 @@
         "nuxt-primevue": "^3.0.0",
         "primevue": "^3.53.0",
         "prismjs": "^1.29.0",
-        "vue": "^3.4.34"
+        "vue": "^3.4.34",
+        "html-truncate": "^1.2.2"
```
hroth1994, line 32: I think `html-truncate` could be added just to `es-ds-components`?

es-ds-docs/pages/molecules/support.vue
```diff
@@ -59,8 +59,8 @@ const supportProps = [
 
         <ds-doc-source
             :comp-code="compCode"
-            comp-source="es-ds-components/components/es-pagination.vue"
+            comp-source="es-ds-components/components/es-support.vue"
```
hroth1994, line 62: Good catch! 🎣

es-ds-docs/pages/molecules/view-more.vue
```diff
@@ -0,0 +1,37 @@
...
+            :content="`Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi in aliquam ex.
+                Nullam vestibulum ex mi, ut suscipit libero condimentum id. Pellentesque eu diam vel nisi
+                molestie porta eget sed odio. Quisque congue risus id metus facilisis, non imperdiet libero rutrum.
+                Mauris vitae ante porttitor, consectetur purus faucibus, euismod ex. Orci varius natoque penatibus
+                et magnis dis parturient montes, nascetur ridiculus mus. Nulla ullamcorper elit sed viverra finibus.
+                Mauris vitae tortor mauris. Cras suscipit nibh nec nisi cursus ornare. Maecenas quis turpis sit amet
+                sapien dapibus sollicitudin viverra eu justo. Vivamus posuere metus sit amet purus tempus volutpat.
+                Donec eleifend elit quam. `" />
+
+        <ds-doc-source
```
hroth1994, line 31: I think this component could use a props table

es-ds-docs/pages/molecules/view-more.vue
```diff
@@ -0,0 +1,37 @@
...
+    compCode.value = $prism.normalizeCode(compSource.default);
+    docCode.value = $prism.normalizeCode(docSource.default);
+    $prism.highlight();
+}
+</script>
+
+<template>
+    <div>
+        <h1>View more</h1>
+
```
hroth1994, line 18: Could include a link to html-truncate - `Uses html-truncate`

hroth1994 (commented):
Looks great so far! Just a few non-functional comments:

es-ds-docs/pages/molecules/view-more.vue
lgeggleston: Forgot about that, thanks!

es-ds-docs/package.json
lgeggleston: True, removed it from `es-ds-docs`

es-ds-docs/pages/molecules/view-more.vue
```diff
@@ -0,0 +1,85 @@
...
+<template>
+    <div>
+        <h1>View more</h1>
+        <p>
+            Using
+            <nuxt-link
+                href="https://www.npmjs.com/package/html-truncate"
+                target="_blank">
+                html-truncate </nuxt-link
```
hroth1994, line 58: I would remove the extra space

hroth1994 (approved):
I agree adding a note to the EsReview ticket is sufficient. One small nit but LGTM! 🌟

PR #1524, opened 2024-09-11
feat: v3 EsSupportCard

hroth1994:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1786

### ❓ Type of change

### 📚 Description

Create EsSupportCard with the [same props API](https://design.energysage.dev/organisms/es-support-card/) for backwards compatibility. Docs page uses `fullWidthComponent` layout instead of default, and the props and slots tables now use `ds-prop-table`.

### 🥼 Testing

Tested locally on http://localhost:8500/organisms/support-card

- General

### 📝 Checklist

### DIFF
--- a/es-ds-components/components/es-support-card.vue
+++ b/es-ds-components/components/es-support-card.vue
@@ -1,0 +1,150 @@
+<script setup lang="ts">
+const props = defineProps({
+    constrained: {
+        type: Boolean,
+        default: false,
+    },
+    imageAltText: {
+        type: String,
+        default: '',
+    },
+    imageSrc: {
+        type: String,
+        default: '',
+    },
+    primaryCtaTarget: {
+        type: String,
+        default: '_blank',
+    },
+    primaryCtaText: {
+        type: String,
+        required: true,
+    },
+    primaryCtaUrl: {
+        type: String,
+        required: true,
+    },
+    secondaryCtaTarget: {
+        type: String,
+        default: '_blank',
+    },
+    secondaryCtaText: {
+        type: String,
+        default: '',
+    },
+    secondaryCtaUrl: {
+        type: String,
+        default: '',
+    },
+});
+
+const showSecondaryCta = computed(() => {
+    return !!props.secondaryCtaText && !!props.secondaryCtaUrl;
+});
+</script>
+
+<template>
+    <es-card
+        class="p-100 px-sm-200 text-center text-lg-left"
+        :class="{
+            'px-md-100 py-md-200': constrained,
+            'px-md-300': !constrained,
+        }">
+        <b-row class="align-items-lg-center justify-content-xl-between">
+            <b-col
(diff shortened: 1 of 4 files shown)


es-ds-components/components/es-support-card.vue
```diff
@@ -0,0 +1,150 @@
...
+            </b-col>
+        </b-row>
+    </es-card>
+</template>
+
+<style lang="scss">
+@use '@energysage/es-ds-styles/scss/mixins/breakpoints' as breakpoints;
+
+@include breakpoints.media-breakpoint-up(lg) {
+    .EsSupportCard {
```
ericdouglaspratt, line 128: Can we move these styles to a SASS file in es-ds-styles so all component styles live in there? Perhaps in the `/scss/vue/components/` folder.

es-ds-docs/components/ds-prop-table.vue
```diff
@@ -34,7 +34,11 @@ export default {
                     {{ column }}
                 </template>
                 <template #value>
-                    <code v-if="columnIndex < 3 && row[columnIndex] !== 'n/a'">
+                    <code
+                        v-if="
+                            ((widths.md.length > 3 && columnIndex < 3) || columnIndex < 2) &&
```
ericdouglaspratt, line 39: Can we add a comment explaining what this new logic does?

ericdouglaspratt (changes requested):
Nice work! A couple things.

es-ds-components/components/es-support-card.vue
nathanielwarner: We haven't really standardized this, but I do think there's an argument for keeping them in the component as well, particularly if we're able to make it `scoped`. It will reduce the size of the CSS bundle that gets served on every page.

nathanielwarner: I've generally kept them in the component unless there's a ton of styles and I'm not able to make it scoped

ericdouglaspratt: That's interesting, and worth a discussion. If we want the bundle to be smaller, that'd end up with us putting as much of the component CSS in the component files themselves, which would eventually result in es-ds-styles only containing global styles.

Since it looks like we have inconsistency already on this, @hroth1994 I'd say it's fine to leave the way you have it.

nathanielwarner: > If we want the bundle to be smaller, that'd end up with us putting as much of the component CSS in the component files themselves, which would eventually result in es-ds-styles only containing global styles.

Yeah honestly the more I think about it, the more this seems like the right approach. We can reduce the global CSS bundle size, and prevent people from relying on CSS classes that are actually tailored to a specific component implementation and subject to change.

ericdouglaspratt (approved):

PR #1526, opened 2024-09-12
feat: Ced 1760 create v3 es verification code using prime vue

RANE1021:
### 🔗 Linked issue
- https://energysage.atlassian.net/browse/CED-1760

### ❓ Type of change

- props : should length be included although it is hard coded?
  - what other props would this need? (IDK it needs other but unsure)
- emits : this was being used to mimic submission to the server, still needed?

### 📝 Checklist


RANE1021:
@hroth1994 or @lgeggleston mind answering my questions above?

hroth1994:
```
props : should length be included although it is hard coded?
```
Length probably shouldn't be hardcoded, as it was a prop in the [existing implementation](https://github.com/EnergySage/es-ds/blob/main/es-vue-base/src/lib-components/EsVerificationCode.vue) (previously charCount)
```
what other props would this need? (IDK it needs other but unsure)
```
See above, it should include all the props from the existing implementation.
```
emits : this was being used to mimic submission to the server, still needed?
``` 
This is definitely necessary, the mimicked submission to the server is just on the docs side

es-ds-docs/components/ds-molecules-list.vue
```diff
@@ -72,5 +72,8 @@
         <li>
             <ds-link to="/molecules/view-more"> View more </ds-link>
         </li>
+        <li>
+            <ds-link to="/molecules/verification-code"> Verification Code </ds-link>
+        </li>
```
ericdouglaspratt, line 77: Nit: Alphabetically, this should come before "View more" in the list.

Also: Can we use sentence case to be consistent with the other list items, so it's "Verification code" rather than "Verification Code"?

es-ds-docs/pages/molecules/verification-code.vue
```diff
@@ -0,0 +1,124 @@
...
+    const docSource = await import('./verification-code.vue?raw');
+    /* eslint-enable import/no-webpack-loader-syntax, import/no-self-import */
+
+    compCode.value = $prism.normalizeCode(compSource.default);
+    docCode.value = $prism.normalizeCode(docSource.default);
+    $prism.highlight();
+}
+
+const propTableRows = [
+    ['v-model', 'String', 'n/a', 'Required. The v-model directive binds the otpinput to a data property.'],
```
ericdouglaspratt, line 18: `charCount` is also a prop we should document here

Also, we can probably just say "input" instead of "otpinput", as "otpinput" is an internal implementation detail of this component that consumers of this component won't have to worry about.

es-ds-docs/pages/molecules/verification-code.vue
```diff
@@ -0,0 +1,124 @@
...
+    }, 1000);
+    return rand.join('');
+};
+
+// eslint-disable-next-line no-return-assign
+const clearCode = () => {
+    isSubmitInProgress.value = true;
+    setTimeout(() => {
+        isSubmitInProgress.value = false;
+    }, 1000);
```
ericdouglaspratt, line 50: On the [current design system page](https://design.energysage.dev/molecules/es-verification-code), the clear code button executes immediately without a simulated server delay, so I think we can remove this `setTimeout` here. Clearing a code is just something that would happen client-side and wouldn't need server interaction.

es-ds-docs/pages/molecules/verification-code.vue
```diff
@@ -0,0 +1,124 @@
...
+    }
+    return false;
+};
+
+const randomCode = () => {
+    // Randomly generates an array of numbers 0-9
+    const rand = Array.from({ length: charCount }, () => Math.floor(Math.random() * 10).toString());
+
+    // eslint-disable-next-line no-console
+    console.log('Generated', rand);
```
ericdouglaspratt, line 37: This `console.log` may have existed on the previous docs page, but since these docs pages should be best practice examples of how to code in Nuxt/Vue, I don't think it's appropriate to include a `console.log` here in an example; I'd say it's safe to remove this.

es-ds-components/components/es-verification-code.vue
```diff
@@ -0,0 +1,89 @@
...
+const updateCode = (newValue: any) => {
+    model.value = newValue;
+};
+</script>
+
+<template>
+    <div>
+        <input-otp
+            v-model="verificationCode"
+            class="custom-otp"
```
ericdouglaspratt, line 29: On the docs page, it looks like you're passing `isSubmitInProgress` in through a `disabled` prop to this component, but then not doing anything with it, so the inputs don't get disabled during the simulated server request. If you add a `v-bind="$attrs"` here on `<input-otp>`, that should pass anything else through, including the `disabled` prop.

es-ds-docs/pages/molecules/verification-code.vue
```diff
@@ -0,0 +1,124 @@
...
+        isSubmitInProgress.value = false;
+    }, 1000);
+    return '';
+};
+</script>
+
+<template>
+    <div>
+        <h1>Verification code</h1>
+        <p>{{ code }}</p>
```
ericdouglaspratt, line 58: We probably don't need this, as the code is displayed in the inputs themselves, and the old docs page didn't have this.

es-ds-docs/pages/molecules/verification-code.vue
```diff
@@ -0,0 +1,124 @@
...
+            <es-verification-code
+                v-model="code"
+                :char-count="charCount"
+                :disabled="isSubmitInProgress" />
+        </b-row>
+        <b-row class="mt-200 align-items-center">
+            <b-col
+                class="mb-100 mb-lg-0"
+                cols="10"
+                lg="4">
```
ericdouglaspratt, line 77: Due to our new design system docs pages more closely adhering to our content container max widths (unlike the old ones), this section is a bit width-constrained on most breakpoints and wraps awkwardly.

Screen Shot 2024-10-02 at 5 14 31 PM

Can you finesse the columns here at various breakpoints so the buttons either wrap to the next line fully underneath the text, or have enough room to go side-by-side with the text without one button wrapping underneath the other?

es-ds-components/components/es-verification-code.vue
```diff
@@ -0,0 +1,89 @@
...
+
+<style lang="scss" scoped>
+@use '@energysage/es-ds-styles/scss/variables' as variables;
+@use '@energysage/es-ds-styles/scss/mixins/breakpoints' as breakpoints;
+
+.custom-otp {
+    display: flex;
+    gap: 10px;
+}
+.custom-otp-input {
```
ericdouglaspratt, line 53: If we're moving this CSS here from `es-ds-styles/scss/vue/components/_verification-code.scss` so that it doesn't get unnecessarily included in an app unless the app uses this component, we should probably also delete that `_verification-code.scss` file so we can actually realize those file size savings.

es-ds-components/components/es-verification-code.vue
```diff
@@ -0,0 +1,89 @@
...
+    },
+});
+
+const count = ref(props.charCount);
+watch(model, (newVal) => {
+    verificationCode.value = newVal;
+});
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const updateCode = (newValue: any) => {
```
ericdouglaspratt, line 20: Weird, I can see why you did this `any` here. I tried changing this type to string, but then `@update:model-value="updateCode"` complained that it was expecting a function that takes a boolean argument, which makes no sense, since this would come in as a string.

So that future generations don't get tripped up by this, can we add a comment here explaining that `@update:model-value` is erroneously requiring a `boolean` function argument, so we've worked around this by specifying `any`?

es-ds-components/components/es-verification-code.vue
```diff
@@ -0,0 +1,89 @@
...
+    &::-webkit-outer-spin-button,
+    &::-webkit-inner-spin-button {
+        appearance: none;
+    }
+
+    &:last-child {
+        margin-right: 0 !important;
+    }
+}
+.red-input {
```
ericdouglaspratt, line 86: It doesn't look like this class is used - can we remove?

es-ds-docs/pages/molecules/verification-code.vue
```diff
@@ -0,0 +1,124 @@
...
+                    :disabled="isSubmitInProgress"
+                    @click="code = clearCode()">
+                    clear code
+                </es-button>
+            </b-col>
+        </b-row>
+    </div>
+    <div>
+        <div class="mb-500">
+            <h2>EsVerification props</h2>
```
ericdouglaspratt, line 114: Typo: should be "EsVerificationCode" rather than "EsVerification"

ericdouglaspratt (changes requested):
Nice start! A few things, plus the below item.

Can you add `primevue/inputotp` to the list of NPM libraries in es-ds-components/nuxt.config.js, so that we don't get reload behavior like this:
```
ℹ Vite server warmed up in 2530ms
ℹ ✨ new dependencies optimized: primevue/inputotp
ℹ ✨ optimized dependencies changed. reloading
```

es-ds-docs/pages/molecules/verification-code.vue
```diff
@@ -0,0 +1,124 @@
+<script setup lang="ts">
+const { $prism } = useNuxtApp();
+const compCode = ref('');
+const docCode = ref('');
+
+if ($prism) {
```
nathanielwarner, line 6: Need to wrap this whole block in an async function inside onMounted as has been done in other components, for prism to work correctly

ericdouglaspratt (changes requested):
Nice improvements! A couple items remain.

es-ds-docs/pages/molecules/verification-code.vue
```diff
@@ -0,0 +1,120 @@
...
+                <es-button
+                    class="mr-1"
+                    :disabled="isSubmitInProgress"
+                    @click="code = randomCode()">
+                    random code
+                </es-button>
+                <es-button
+                    outline
+                    :disabled="isSubmitInProgress"
+                    @click="code = clearCode()">
```
ericdouglaspratt, line 102: For simplicity, now that this function is just returning an array, this can probably just be `code = []` and we can remove the function.

ericdouglaspratt (changes requested):
Looks good, but I'm seeing these errors now in the console:
```
[Vue warn]: Failed to resolve component: b-row
[Vue warn]: Failed to resolve component: b-col
```

Tom's PR to rename `b-row` to `es-row` and `b-col` to `es-col` has been merged, so you'll want to do a search and replace to take care of those name updates. That's why the layout looks a little broken right now, with the random/clear buttons always being below the code printout text.

ericdouglaspratt (approved):
Nice!

PR #1529, opened 2024-09-16
CED-1790 create v3 EsZipCodeForm

lgeggleston:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1790

### ❓ Type of change

### 📚 Description

- Recreated EsZipCodeForm for vue 3
- Note: One change from Vue 2 showed up in `EsFormInput` - `inheritAttrs: false` [previously did not include classes but does now](https://vueframework.com/guide/migration/attrs-includes-class-style.html#overview), and the zip code form needs to pass some classes to the input-wrapper.  I added a workaround to enable in this case
- Note: Added `vuelidateInteger`, because the updated version now allows decimals with `vuelidateNumber` which does make sense, this is more specific

### 🥼 Testing

- Tested watcher on `zipCodeValue` via a timeout to set it on the docs page
- basic a11y testing for keyboard navigation and aria attributes

- I removed the class `invalid`, because it didn't seem to have an effect or be styled anywhere now, perhaps replaced by `is-invalid` on form input.  Would love a 2nd opinion on this however

### 📝 Checklist


tomleo:
👀

tomleo:
This isn't a regression as this behavior matches what we had before, but the side-by-side doesn't seem to be optimal in mobile

Screenshot 2024-09-16 at 10 24 45 AM

es-ds-components/components/es-zip-code-form.vue
```diff
@@ -0,0 +1,204 @@
...
+        type: String,
+        default: '',
+    },
+    selectedProduct: {
+        type: String,
+        default: '',
+    },
+});
+
+const state = reactive({
```
tomleo, line 57: nit: I'd recommend using `ref`, mostly for consistency (it should result in the same thing in this context)

es-ds-components/components/es-zip-code-form.vue
```diff
@@ -0,0 +1,204 @@
...
+    zipCode: props.zipCodeValue,
+});
+
+watch(
```
tomleo, line 61: It looks like you're binding `state.zipCode` to the input, watching for that change, and updating the state accordingly.

I wonder if using:

```
const model = defineModel()
```

Then setting `v-mode=model` would be more optimal here

tomleo (commented):
Overall looks good 👍

One possible tweak with v-model on the input

es-ds-components/components/es-zip-code-form.vue
lgeggleston: Consistency with what? I used `state` to match the new[ form validation example](https://new.design.energysage.dev/examples/form-field-validation) as much as possible; 
I do agree that `ref` should probably work the same way for a single value, but when I tried that with `useEsForms` the validation was broken so I'm not sure it's quite supported to not wrap it in `state`.

es-ds-components/components/es-zip-code-form.vue
lgeggleston: You're right that setting this up as a `v-model` might have made sense for this too (although may get into messiness interacting with the validation setup above) - but I would rather not make that change, since it'd be an API change for downstreams and avoidable.  I don't remember exactly why it was originally set up this way ([PR for context](https://github.com/EnergySage/es-ds/pull/1385)), but it only needs one-way data binding as a prop rather than 2-way binding w upstream component.

es-ds-components/components/es-zip-code-form.vue
nathanielwarner: `ref` works differently since the value is accessed through `.value` rather than on the object itself. Vuelidate requires using either `reactive` or `computed` for the form state.

nathanielwarner: (Actually, you can also use an object of refs, like `{ zipCode: ref(props.zipCodeValue) }`. But I think using `reactive` looks cleaner and aligns with the form validation example. [Their docs for reference](https://vuelidate-next.netlify.app/advanced_usage.html#composition-api))

lgeggleston:
> This isn't a regression as this behavior matches what we had before, but the side-by-side doesn't seem to be optimal in mobile

That's a fair point for this example in the docs page, but I'm going to say this is a use case that should still be supported.  The `stackUntil` just allows any viewport to be entered so it's up to the instance to set it at a viewport that makes sense; and there's a possibility it can look fine side-by-side on mobile if someone wants a short button text:
Screenshot 2024-09-18 at 11 49 14 AM
I updated the docs page button to that so that it makes more sense.

es-ds-components/components/es-zip-code-form.vue
lgeggleston: gotcha, thanks for the context!

tomleo (approved):
👍

PR #1533, opened 2024-09-19
feat: add search bar to nav bar

hroth1994:
### ❓ Type of change

### 📚 Description

Adds a search bar component to the nav bar, currently only for desktop views. 
Adds a prop to the nav bar `showSearch` to enable the search icon. Defaults to `false` as to not affect existing functionality.

### 🥼 Testing

Tested locally at http://localhost:8500/organisms/es-nav-bar - click search icon to open/close, x icon also closes search bar
Screenshot 2024-09-19 at 4 18 25 PM

- General

### 📝 Checklist

### DIFF
--- a/es-vue-base/src/lib-components/EsNavBar.vue
+++ b/es-vue-base/src/lib-components/EsNavBar.vue
@@ -120,7 +120,8 @@
                     <!-- mobile+desktop product menus -->
                     <b-container v-if="searchBarOpen">
                         <div class="row w-100">
-                            <es-nav-bar-search-bar>
+                            <es-nav-bar-search-bar
+                                v-model="searchText">
                                 <template #close>
                                     <es-button
                                         class="position-absolute"
@@ -283,6 +284,7 @@ export default {
     data() {
         return {
             searchBarOpen: false,
+            searchText: '',
         }
     },
     methods: {
--- a/es-vue-base/src/lib-components/EsNavBar.vue
+++ b/es-vue-base/src/lib-components/EsNavBar.vue
@@ -288,7 +288,6 @@ export default {
     methods: {
         toggleSearchBar() {
             this.searchBarOpen = !this.searchBarOpen;
-            console.log(this.searchBarOpen);
         }
     },
     mounted() {
--- a/es-vue-base/src/lib-components/EsNavBar.vue
+++ b/es-vue-base/src/lib-components/EsNavBar.vue
@@ -100,6 +100,17 @@
                                 <slot name="logo" />
                             </template>
                         </es-nav-bar-top-level-menu>
+                        <div v-if="showSearch" class="nav-item d-none d-lg-block pt-100">
+                            <es-button
+                                variant="link"
+                                class="nav-link dropdown-toggle d-none d-lg-flex flex-nowrap py-100"
+                                @click="toggleSearchBar()">
+                                <icon-search
+                                    class="align-self-center account-icon"
+                                    width="20px"
+                                    height="20px" />
+                            </es-button>
+                        </div>
                         <!-- desktop account menu -->
                         <es-nav-bar-account-menu
                             :auth-items="accountContent.loggedIn.items"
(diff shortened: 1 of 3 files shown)


es-vue-base/src/lib-components/EsNavBar.vue
```diff
@@ -100,14 +100,48 @@
                                 <slot name="logo" />
                             </template>
                         </es-nav-bar-top-level-menu>
+                        <div
+                            v-if="showSearch"
+                            class="nav-item d-none d-lg-block pt-100">
+                            <es-button
```
nathanielwarner, line 106: This button needs a focus state so users navigating by keyboard can see when they're on it

nathanielwarner: It also needs an `aria-label`

es-vue-base/src/lib-components/EsNavBar.vue
```diff
@@ -100,14 +100,48 @@
...
+                        <div class="row w-100">
+                            <es-nav-bar-search-bar
+                                v-bind="$attrs"
+                                v-on="$listeners">
+                                <template #close>
+                                    <es-button
+                                        class="position-absolute"
+                                        style="right: 0"
+                                        variant="link"
+                                        @click="toggleSearchBar()">
```
nathanielwarner, line 133: Same feedback for this button

es-vue-base/src/lib-components/EsNavBarSearchBar.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+    <div class="align-items-center w-100 d-flex justify-content-center position-relative">
+        <div class="d-flex align-items-center mx-auto justify-content-center w-100">
+            <es-form-input
```
nathanielwarner, line 4: With ESDS v2, I believe design system components including `EsFormInput` and `EsButton` and `IconSearch` need to be imported to be usable in downstream repos, even though they work in the docs

es-vue-base/src/lib-components/EsNavBarSearchBar.vue
```diff
@@ -0,0 +1,38 @@
...
+                class="mt-3 w-50"
+                v-bind="$attrs"
+                label-sr-only
+                :placeholder="label"
+                v-on="$listeners">
+                <template #prefixIcon>
+                    <icon-search />
+                </template>
+            </es-form-input>
+            <es-button class="ml-50">
```
nathanielwarner, line 15: As of right now, this button can't do anything. I think the easiest way to wire things up from how you have it now is to emit an event from this button (say "searchButtonClicked"), which you can then listen to from the `nav` layout in the docs (or whatever downstream is implementing this). This listener will be responsible for navigating the user to the search results page, most likely passing query parameters which will allow the necessary API call to Algolia to be made. But all you have to worry about in the design system is verifying that the event works, so you could just do a `console.log` or something.

nathanielwarner (changes requested):
Good start on this!

es-vue-base/src/lib-components/EsNavBar.vue
```diff
@@ -100,14 +100,51 @@
...
                             :logged-out="accountContent.loggedOut" />
                     </b-container>
                     <!-- mobile+desktop product menus -->
-                    <b-container class="d-flex flex-lg-nowrap justify-content-lg-end product-menu">
+                    <b-container v-if="searchBarOpen">
+                        <div class="row w-100">
+                            <es-nav-bar-search-bar
+                                v-bind="$attrs"
+                                v-on="$listeners"
+                                @searchButtonClicked="searchButtonClicked">
```
nathanielwarner, line 129: Because we already have `v-on="$listeners"`, we don't need `@searchButtonClicked="searchButtonClicked"` here, and having it results in the event being fired twice.

es-vue-base/src/lib-components/EsNavBarSearchBar.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+    <div class="align-items-center w-100 d-flex justify-content-center position-relative">
+        <div class="d-flex align-items-center mx-auto justify-content-center w-100">
+            <es-form-input
+                id="searchBar"
+                class="mt-3 w-50"
+                v-bind="$attrs"
+                label-sr-only
+                :placeholder="label"
+                v-on="$listeners">
```
nathanielwarner, line 10: Right now clicking "enter" doesn't submit the form, which I imagine is something we want.. We can fix easily by emitting the event when that happens as well

PR #1540, opened 2024-10-01
CED-1761

tomleo:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-1761

### ❓ Type of change

### 📚 Description

- port `EsVideo` component to vue3

### 🥼 Testing

- manual

- I had a working auto-play on click solution via https://github.com/EnergySage/es-ds/commit/397755e3046d7210cea42a5b1062745935c3ee8d, but the player would need to live outside the component. I think a composable on a page featuring videos would be the right way to handle that.
- There's a CLS when the palceholder image loads, and the dimensions are fluid. I opted for calculating the size of the card when clicked on so that the iFrame will have the exact size. All this logic could be simplified if we had fixed sizes, but that might be something better suited for higher-order components. 

### 📝 Checklist


es-ds-components/components/es-video.vue
```diff
@@ -0,0 +1,73 @@
...
+import { useElementSize } from '@vueuse/core';
+
+interface IProps {
+    altText: string;
+    coverImageUrl: string;
+    embedUrl: string;
+}
+const props = defineProps<IProps>();
+const showVideo = ref(false);
+function getVideoIdFromUrl(url: string) {
```
ericdouglaspratt, line 11: Nit: this is all kind of jammed together - thoughts on adding blank lines to separate semantic sections from each other? e.g. props/ref definitions from functions here, functions from const variables below, etc.

es-ds-components/components/es-video.vue
```diff
@@ -0,0 +1,73 @@
...
+    const embedPattern = /https:\/\/www\.youtube\.com\/embed\/(?<videoId>[A-Za-z0-9-]+)/;
+    const parsedUrl = embedPattern.exec(url);
+    if (parsedUrl) {
+        return parsedUrl?.groups?.videoId;
+    }
+    return '';
+}
+const videoId = getVideoIdFromUrl(props.embedUrl);
+const embedUrlWithParams = `https://www.youtube.com/embed/${videoId}?rel=0&autoplay=1&cc_load_policy=1&cc_lang_pref=en`;
```
ericdouglaspratt, line 20: Because these two variables depend on props, I think we may need to make them [computed](https://vuejs.org/guide/essentials/computed.html) so they can be reactive and react if the prop values change.

es-ds-components/components/es-video.vue
```diff
@@ -0,0 +1,73 @@
...
+
+const placeholderCardEl = useTemplateRef('placeholder-card');
+const cardDimensionElementSize: { width: number; height: number } = reactive(
+    useElementSize(placeholderCardEl, { width: 0, height: 0 }, { box: 'border-box' }),
+);
```
ericdouglaspratt, line 25: At the moment, when the iframe is loaded on mobile, it stops retaining its aspect ratio if the window is resized, whereas the old EsVideo did retain its aspect ratio. I think we should try to keep this behavior.

The `<b-embed>` component uses a few simple CSS classes to handle keeping the 16:9 ratio we want; I think we should copy those over and use those as a simpler solution than live-listening to the size of the card and updating a bunch of JS variables.

These CSS classes already appear in our es-ds-styles in the [embed utilities SASS file](https://github.com/EnergySage/es-ds/blob/esds-3.0-vue3-primevue/es-ds-styles/scss/utilities/_embed.scss), so they're already there for us to use. We just need to do [what <b-embed> does](https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/components/embed/embed.js#L39) and put the associated classes on the parent element of the iframe.

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,64 @@
...
+    }
+});
+</script>
+
+<template>
+    <div>
+        <h1>Video</h1>
+        <div class="my-500">
+            <b-row>
+                <!-- TODO: set explicit width & height on on es-video container to avoid CLS -->
```
ericdouglaspratt, line 29: Switching to the CSS method described above may also solve for this

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,64 @@
...
+        <div class="mb-500">
+            <h2>Props</h2>
+            <ds-prop-table :rows="propTableRows" />
+        </div>
+
+        <ds-doc-source
+            :comp-code="compCode"
+            comp-source="es-ds-components/components/es-form-input.vue"
+            :doc-code="docCode"
+            doc-source="es-ds-docs/pages/molecules/text-input.vue" />
```
ericdouglaspratt, line 62: Need to update these from text input to video

es-ds-styles/scss/utilities/_position.scss
```diff
@@ -32,3 +32,11 @@
     z-index: variables.$zindex-sticky;
   }
 }
+
+// Used with position-absolute to center an element within it's container
+.abs-center {
+    left: 50%;
+    top: 50%;
+    transform: translateX(-50%) translateY(-50%);
+    z-index: variables.$zindex-overlay;
```
ericdouglaspratt, line 41: Hmm 1035 is pretty high - I _think_ that because this is within the position relative context of the container, that forms a separate stacking context from all other z-indices on the page, meaning the play icon might not randomly appear on top of a modal, for example, but I'm not 100% sure.

ericdouglaspratt (changes requested):

es-ds-components/components/es-video.vue
```diff
@@ -0,0 +1,73 @@
...
+        </div>
+        <es-card
+            v-else
+            :id="`youtube-placeholder-${videoId}`"
+            ref="placeholder-card"
+            class="EsVideo-button bg-transparent overflow-hidden p-0 position-relative w-100"
+            variant="interactive"
+            @click="handleLoadVideo">
+            <slot v-if="$slots.default" />
+            <img
```
hroth1994, line 61: Could use `nuxt-img` here

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,64 @@
+<script setup lang="ts">
+const propTableRows = [
+    ['alt-text', 'String', 'n/a', 'Required.'],
```
hroth1994, line 3: These props need a complete description, what would someone using `EsVideo` for the first time need to know?

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,64 @@
...
+            <b-row>
+                <!-- TODO: set explicit width & height on on es-video container to avoid CLS -->
+                <b-col md="8">
+                    <es-video
+                        :alt-text="`
+                            ME from EnergySage talking to you about how to decide if solar panels
+                            are worth it for your needs
+                        `"
+                        cover-image-url="https://a-us.storyblok.com/f/1006159/1280x720/2cb0626421/are-solar-panels-worth-it_.jpg/m/810x0/filters:format(jpg):quality(70)"
+                        embed-url="https://www.youtube.com/embed/srBDwFZqOi0" />
```
hroth1994, line 37: Do we want to use real ES videos? Could stick with the cat theme here

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,64 @@
...
+
+<template>
+    <div>
+        <h1>Video</h1>
+        <div class="my-500">
+            <b-row>
+                <!-- TODO: set explicit width & height on on es-video container to avoid CLS -->
+                <b-col md="8">
+                    <es-video
+                        :alt-text="`
```
hroth1994, line 32: This can be a string, and doesn't need to be in backticks

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,64 @@
...
+        </div>
+
+        <div class="mb-500">
+            <h2>Props</h2>
+            <ds-prop-table :rows="propTableRows" />
+        </div>
+
+        <ds-doc-source
+            :comp-code="compCode"
+            comp-source="es-ds-components/components/es-form-input.vue"
```
hroth1994, line 60: Should use es-video for `comp-source` and `doc-source`

es-ds-styles/scss/utilities/_position.scss
```diff
@@ -32,3 +32,11 @@
     z-index: variables.$zindex-sticky;
   }
 }
+
+// Used with position-absolute to center an element within it's container
```
hroth1994, line 36: Cool!

hroth1994 (commented):
Looking good so far! A few comments:

es-ds-components/components/es-video.vue
tomleo: Ahh yeah CSS solution is much better 👍

es-ds-styles/scss/utilities/_position.scss
tomleo: I decoupled the z-index from `abs-center` and added a modal prop to `es-video`. I think this addresses the issue if I understand correctly.

tomleo: A couple quick tests, and it looks like things behave as expected without using `variables.$z-index-modal-overlay`, going to revert that bit.

Screenshot 2024-10-03 at 12 40 39 PM

Screenshot 2024-10-03 at 12 40 55 PM

hroth1994: Yea that's why I ended up using https://loremflickr.com/200/200?random=3 🙁

es-ds-docs/pages/molecules/video.vue
hroth1994: Yea that's why I ended up using https://loremflickr.com/200/200?random=3 🙁

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,86 @@
...
+                        alt-text="The worst thing about prison was the Dementors."
+                        cover-image-url="/placeholders/810-456.jpg"
+                        embed-url="https://www.youtube.com/embed/djV11Xbc914" />
+                </b-col>
+            </b-row>
+        </div>
+
+        <div class="my-500">
+            <h2>Video in modal</h2>
+            <button @click="showModal = !showModal">Show/Hide Modal</button>
```
hroth1994, line 58: nit: sentence case

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,86 @@
...
+            <es-modal
+                id="simple-modal"
+                hide-footer
+                :visible="showModal"
+                size="lg"
+                @hidden="showModal = false">
+                <template #modal-title> Modal Example </template>
```
hroth1994, line 65: nit: sentence case

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,86 @@
...
+                    <es-video
+                        alt-text="Bears, beets, Battlestar Galactica."
+                        cover-image-url="/placeholders/810-455.jpg"
+                        embed-url="https://www.youtube.com/embed/dQw4w9WgXcQ" />
+                </template>
+            </es-modal>
+        </div>
+
+        <div class="mb-500">
+            <h2>Props</h2>
```
hroth1994, line 76: nit: EsVideo props

es-ds-docs/pages/molecules/video.vue
ericdouglaspratt: IMO we should use real-world EnergySage use cases whenever possible to quickly make sense to our internal users of the design system, without explanation, what the purpose of a component is - and when doing so wouldn't compromise any internal information, and when such an example is already public, as our YouTube videos are.

hroth1994 (approved):
3 nits then LGTM! 🌟

es-ds-docs/pages/molecules/video.vue
hroth1994: Could be something to discuss in the design guild, I had thought we were moving away from real ES examples but could be wrong

tomleo: @hroth1994 Totally agree. In the interim though, I reverted back to what was there before as the ticket was to upgrade the component for vue3.

tomleo:
@ericdouglaspratt I think I've addressed your feedback, LMK

ericdouglaspratt (changes requested):
CSS solution looks good! A couple items.

Also, if we're not using them in this PR anymore, can we remove the bear images from the PR?

es-ds-docs/pages/molecules/video.vue
```diff
@@ -0,0 +1,89 @@
+<script setup lang="ts">
+const propTableRows = [
+    ['alt-text', 'String', 'n/a', 'Required. A short description of the video'],
+    ['cover-image-url', 'String', 'n/a', 'Required. The image shown before loading the video'],
+    [
+        'embed-url',
+        'String',
+        'n/a',
+        'Required. The embed URL following the format https://www.youtube.com/embed/:video-id',
```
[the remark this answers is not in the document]
tomleo: I thought that was fairly common, but maybe specific to expressjs 🤔, swapped for `[video id]`

es-ds-styles/scss/_variables.scss
```diff
@@ -1325,6 +1325,7 @@ $form-validation-states: map.merge(
 $zindex-dropdown:                   1000 !default;
 $zindex-sticky:                     1020 !default;
 $zindex-fixed:                      1030 !default;
+$zindex-overlay:                    1035 !default;
```
ericdouglaspratt, line 1328: Nit: since we're not using this, we can probably clean this up too

ericdouglaspratt (approved):
Nice work!

PR #1541, opened 2024-10-02
feat: es-form v3

hroth1994:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-1785

### ❓ Type of change

### 📚 Description

Recreates BForm as EsForm without Bootstrap Vue for ESDS v3. This API is fairly minimal, having just the inline and novalidate props in addition to the props that are passed through to the native form.

### 🥼 Testing

Tested locally with inline enabled at http://localhost:8500/organisms/form

- General

### 📝 Checklist


es-ds-components/components/es-form.vue
```diff
@@ -0,0 +1,38 @@
...
+<template>
+    <form
+        :class="{ 'form-inline': props.inline }"
+        v-bind="$attrs">
+        <slot />
+    </form>
+</template>
+
+<style lang="scss">
+.form-inline {
```
ericdouglaspratt, line 19: Hmm I'm a little split on adding these styles here. 

For one, they're redundant with `.form-inline` styling already present in es-ds-styles/scss/_forms.scss. So we should probably only have them in one place. I [don't see any Vue app references](https://github.com/search?q=org%3AEnergySage%20form-inline&type=code) where people do their own `<form>` tag and use the `form-inline` class directly, outside of es-site and dsire-wordpress, which will probably never use this new Nuxt-3-required version of the design system, so it _may_ be safe to remove from that SCSS file in es-ds-styles.

However, it's generally an antipattern to not use `scoped` and especially to write CSS in a component against class names that don't even appear within a component, like `input-wrapper`. If someone uses the `input-wrapper` class elsewhere and happens to have a `form-inline` up in the parent chain, CSS that they may not be expecting would get applied to it.

I'm leaning towards the simplest solution right now being to remove these styles here and just keep them in es-ds-styles. That way it's extra clear that they apply globally, rather than these non-scoped, globally-applicable styles only being included on a page when this component happens to be included (a bit dangerous and fragile, as it's a bit unpredicable).

es-ds-docs/pages/organisms/form.vue
```diff
@@ -0,0 +1,135 @@
+<script setup lang="ts">
+const inline = ref(false);
+
+const propTableRows = [
+    ['inline', 'Boolean', 'false', 'Display labels and form controls on a single horizontal row'],
+    ['novalidate', 'Boolean', 'null', 'Disables browser native HTML5 validation on controls in the form'],
```
ericdouglaspratt, line 6: Since we're not doing anything special to provide this prop relative to all the other attributes a `<form>` tag can take, I might be in favor of removing this here and letting people look up [attributes of the form tag itself](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attributes_for_form_submission) to see what they can do.

ericdouglaspratt (changes requested):
Nice! A couple items.

ericdouglaspratt (approved):
Just one thing to fix but otherwise good to go!

PR #1548, opened 2024-10-07
feat: use v-bind in the style section

tomleo:
DONT merge, just for reference.

Not really in the scope for [CED-1755](https://energysage.atlassian.net/browse/CED-1755), but I think `startingValue` should be made optional, and the default value set to the `v-model` value. It feels slightly odd to set a non-reactive startValue as well as bind to a reactive value via `v-model`.

[CED-1755]: https://energysage.atlassian.net/browse/CED-1755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/es-ds-components/components/es-slider.vue
+++ b/es-ds-components/components/es-slider.vue
@@ -55,9 +55,6 @@ const props = defineProps({
 // allow use of v-model on this component
 const model = defineModel<number | number[]>();
 model.value = props.startingValue;
-function handleSliderUpdate(newVal: number | number[]) {
-    model.value = newVal;
-}
 </script>
 
 <template>
@@ -80,7 +77,7 @@ function handleSliderUpdate(newVal: number | number[]) {
                     class: 'slider-handle',
                 }
             }"
-            @update:model-value="handleSliderUpdate" />
+        />
 
         <div class="d-flex flex-row justify-content-between">
             <span>{{ labelFormatter(min) }}</span>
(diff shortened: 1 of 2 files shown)


es-ds-components/components/es-slider.vue
```diff
@@ -126,7 +129,7 @@ model.value = props.startingValue;
     bottom: 27px;
     box-shadow: 0 1px 6px 0 rgba(34, 38, 51, 0.25);
     color: variables.$white;
-    content: attr(aria-valuenow);
+    content: v-bind("`'${model}'`");
```
tomleo, line 132: The syntax is super-weird because it's evaluating it the same way it would an attribute in the HTML section. But basically the CSS needs `'string content'`, and with *setup* you need to add `' '` (or `" "`) to the *v-bind*.

es-ds-components/components/es-slider.vue
```diff
@@ -47,14 +47,13 @@ const props = defineProps({
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        default: (val: any) => val,
     },
 });
 
 // allow use of v-model on this component
-const model = defineModel<number>();
-
-// set starting value
+const model = defineModel<number | number[]>();
```
tomleo, line 56: Small note, if hooked directly up to prime-vue's slider, then the value can be of type `number | number[]`.

mpleroux (approved):
Thanks for figuring out this crazy `v-bind` syntax.

PR #1552, opened 2024-10-20
feat: initial cut of segmented control using TabMenu

ericdouglaspratt:
### ❓ Type of change

-

### 📝 Checklist

### DIFF
--- a/es-ds-components/components/es-segmented-control.vue
+++ b/es-ds-components/components/es-segmented-control.vue
@@ -1,0 +1,171 @@
+<script setup lang="ts">
+import TabMenu from 'primevue/tabmenu';
+
+interface SegmentedControlItem {
+    label: string;
+}
+
+interface Props {
+    disabled?: boolean;
+    items: SegmentedControlItem[];
+}
+
+const props = withDefaults(defineProps<Props>(), {
+    disabled: false,
+});
+
+// allow use of v-model on this component
+const model = defineModel<number>();
+
+// since the overall component doesn't have a disabled state but
+// individual items do, if disabled is true, disable each item
+const preparedItems = computed(() => {
+    return props.disabled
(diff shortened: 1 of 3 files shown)


es-ds-components/components/es-segmented-control.vue
```diff
@@ -0,0 +1,237 @@
...
+onUnmounted(() => {
+    if (resizeObserver) {
+        resizeObserver.disconnect();
+    }
+});
+</script>
+
+<template>
+    <div
+        v-if="$props.options?.length"
```
nathanielwarner, line 98: Nit: we don't need to use `$props.` or `props.` in the templates when accessing props. It's sufficient to say `v-if="length"`.

es-ds-docs/pages/molecules/segmented-control.vue
```diff
@@ -0,0 +1,155 @@
...
+        label: '3 months',
+    },
+    {
+        label: '1 year',
+    },
+];
+
+const esSegmentedControlProps = [
+    [
+        'ariaLabel',
```
nathanielwarner, line 27: What does this prop correspond to? It seems like it's not being used or referenced anywhere, and the `label` prop is the one that would be the one for people to set.

nathanielwarner (approved):
Looks very slick! Nice work! Mainly just a minor thing about making sure the props documentation lines up with what's available.

es-ds-docs/pages/molecules/segmented-control.vue
ericdouglaspratt: Good catch - `label` was originally called `ariaLabel` but type checking complained that passing in `:aria-label="something"` didn't match `ariaLabel` in camel case, which was really annoying, so I just simplified the name to `label` and forgot to update the props table to match.

PR #1554, opened 2024-10-23
fix: support root-relative URLs in EsButton

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-1900

### ❓ Type of change

### 📚 Description

- Previously, EsButton, when given either an `href` or `to` prop, would pass those to NuxtLink. NuxtLink will externally navigate with a full URL like https://www.google.com, but if provided a root-relative URL like /market/start/, it will attempt to internally navigate and result in a 404, even though that URL does exist on our domain.
- This changes EsButton so that if `href` is provided, we use an `<a>` tag, and if `to` is provided, we use a NuxtLink. This allows downstream applications to decide whether they need an internal or external navigation on a link, rather than the design system forcing one upon them that may not work.

### 🥼 Testing

- Tested locally with es-cms-pages and LandingCtaFullWidthImage


hroth1994 (approved):
LGTM! 🌟

PR #1560, opened 2024-11-19
feat: match EsButton to design

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-2479

### ❓ Type of change

### 📚 Description

- Matches the updated Figma designs for buttons, especially the 1px border rather than 2px
- All focus states will only appear on `:focus-visible` instead of `:focus`, meaning you won't see them showing up unnecessarily on mouse/tap interaction, but you will see them on keyboard tabbing, which is exactly when Chrome shows its default focus outlines
- Moved all focus outlines to be 2px outside of the button so they're all a consistent size with each other across button variants and better match how all other focus outline states for other elements behave
- Updated the primary button focus outline color from a light blue (that wasn't accessible against the white background) to our primary dark blue to meet [WCAG contrast guidelines](https://www.sarasoueidan.com/blog/focus-indicators/#wcag-2.1-and-wcag-2.2-focus-indicator-accessibility-requirements), pending review from the design team on if they want a different color that meets those guidelines (which I'd argue will take long enough that it should be done in a separate ticket/PR)
- Greatly simplified the Button docs page with more straightforward examples to copy from
- Gave non-inline link buttons a darker active state and the press-down animation to match their siblings
- Added more visible bordered focus state to non-inline link buttons to match WCAG criteria

### Updated docs page
screencapture-localhost-8500-molecules-button-2025-07-08-07_37_08

### Focus states for each button type (manually turned on for each via dev tools; won't show like this on docs page unless each one is tabbed to via keyboard)
Screenshot 2025-07-09 at 10 49 10 AM

### 🥼 Testing

- Tested locally

- Overall

### 📝 Checklist


nathanielwarner (commented):
Looks pretty good, no obvious issues with the code! I do have one concern with the changes in focus state- the dark background examples are now harder to see when navigating by keyboard, as compared to the current implementation. That's because the light blue outline is no longer present on focus. Were these changes part of the scope of the design request?

ericdouglaspratt:
@nathanielwarner Good catch - I've now fixed the focus outlines for the `dark-bg` buttons. I went slightly beyond the Figma design for the focus state of the `dark-bg` outline variant to give it an thicker outline, because I felt the border color change from `$white` to `$blue-50` was too subtle to notice.

Also, it looks in the Figma like the glow on the `dark-bg` regular button is back, so I restored it.

nathanielwarner (commented):
Summary of in-person conversation: changes overall are good, but the focus state should be left as is to continue to conform to [accessibility guidelines](https://www.sarasoueidan.com/blog/focus-indicators/#wcag-2.1-and-wcag-2.2-focus-indicator-accessibility-requirements) which require the outline border to be at least 2px wide

nathanielwarner (approved):
Looks better! I think the outline of the primary button still doesn't meet WCAG requirements because of insufficient contrast between the focus indicator and the button, whereas prior to this PR it wasn't meeting the requirements because the outline was too similar to the white background of pages. We should work with design to find an appropriate color and make sure the various requirements are being met.

PR #1562, opened 2024-12-03
fix: v-for handling in EsTabs and EsAccordion

lgeggleston:
### 🔗 Linked issue

Unblocks https://energysage.atlassian.net/browse/CED-1983 and https://energysage.atlassian.net/browse/CED-1986

### ❓ Type of change

### 📚 Description

With the current implementation of Tabs and Accordion, you'll get errors when trying to use a `v-for` to iterate through child elements (`es-tab` and `es-accordion`), which is what many of the downstream usages do.  When we're accessing the children using useSlots() the `v-for` element shows up as one child, but we need its children in turn, so this is just doing some replacement in that case. 

### 🥼 Testing

Tested with the examples on both docs pages; also tested adding a hardcoded child alongside the `v-for` in both examples

- This feels a bit hacky, definitely open to suggestions if there are other methods to try!

### 📝 Checklist

### DIFF
--- a/es-ds-components/components/es-accordion-list.vue
+++ b/es-ds-components/components/es-accordion-list.vue
@@ -17,19 +17,27 @@ const props = withDefaults(defineProps<Props>(), {
 // eslint-disable-next-line vue/require-prop-types
 const model = defineModel();
 
-const children = useSlots().default?.() || [];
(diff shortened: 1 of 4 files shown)


ericdouglaspratt (approved):
Looks good!

PR #1571, opened 2025-02-04
fix: use hardcoded domain to avoid hydration error

mpleroux:
### 🔗 Linked issue

[CED-2130: Hydration error on 404 page in nuxt 3](https://energysage.atlassian.net/browse/CED-2130)

### ❓ Type of change

### 📚 Description

- Use hardcoded domain instead of the value in `PROCESS.ENV`

After PR feedback:

- Added optional String prop for ES domain

### 🥼 Testing

- Linked `es-cms-pages` to `es-ds` for local testing
- Visited dummy page https://localhost:8700/bla/ to get 404 error
- The hydration error message no longer appears in the Dev Tools console

After PR feedback:

- Confirmed new String prop works by adding parameter to `es-error-page` in `error-page.vue`

After this PR is approved:

- [ ] Merge it into feature branch
- [ ] Publish and version `es-ds 3.0.0-alpha.18`?
- [ ] Create PRs to update other repos:
    - [ ] es-storyblok-shared (add support for new prop in `error.vue`?)
    - [ ] es-cms-pages (vue3-upgrade branch)
    - [ ] es-cms-articles (vue3-upgrade branch)
    - [ ] nuxt-wp-redirects

### 📝 Checklist


es-ds-components/components/es-error-page.vue
```diff
@@ -4,7 +4,7 @@ import IconCommunitySolar from './icon/community-solar.vue';
 import IconInfoBlogPost from './icon/info-blog-post.vue';
 import IconMarketplace from './icon/marketplace.vue';
 
-const ES_DOMAIN = process.env.ES_DOMAIN || 'https://www.energysage.com';
+const ES_DOMAIN = 'https://www.energysage.com';
```
ericdouglaspratt, line 7: Should we make this an optional prop instead? Where it still defaults to `https://www.energysage.com` if the prop is not provided? That way the error page can have proper links on int, and the downstream repo can handle getting the appropriate env variable and passing it in.

ericdouglaspratt: Took me a bit of searching to find it too - it's error.vue in storyblok-shared.

es-ds-components/components/es-error-page.vue
```diff
@@ -4,7 +4,21 @@ import IconCommunitySolar from './icon/community-solar.vue';
...
 import IconMarketplace from './icon/marketplace.vue';
 
-const ES_DOMAIN = process.env.ES_DOMAIN || 'https://www.energysage.com';
+const props = defineProps({
+    errorType: {
+        type: String,
+        required: true,
+    },
+    useEsDomain: {
+        type: Boolean,
```
ericdouglaspratt, line 13: What I meant was an optional string, maybe `esDomain` that would be used directly, and then we don't use `process.env.ES_DOMAIN` at all, we use `props.esDomain` directly, or the energysage.com string as a backup.

mpleroux: I changed it to a optional string prop.

I can see the appropriate URL in the links when I test the new parameter on the docs page http://localhost:8500/organisms/error-page

es-ds-components/components/es-error-page.vue
```diff
@@ -4,7 +4,19 @@ import IconCommunitySolar from './icon/community-solar.vue';
...
+        type: String,
+        required: true,
+    },
+    esDomain: {
+        type: String,
+        default: null,
+    },
+});
+
+const ES_DOMAIN = props.esDomain || 'https://www.energysage.com';
```
ericdouglaspratt, line 18: One thing I realized during testing - if I load the error page docs page locally, then pass in a value to the `esDomain` prop, the change is not reflected.

This is because the `ES_DOMAIN` variable and the `linkItems` variable that depends on it are not set up to be reactive to props changes. Now that we have a prop, everything that flows from that prop should generally always be reactive in a Vue component.

This is easy to do if we just wrap a `computed(() => { ... })` around the logic for both `ES_DOMAIN` (which at that point can probably become a camelCase name since it's no longer a fixed constant) and `linkItems`.

ericdouglaspratt (changes requested):

es-ds-components/components/es-error-page.vue
mpleroux: I made those changes and I still see the correct domain when I pass `es-domain` as a test in `error-page.vue`.

ericdouglaspratt (approved):
Looks good!

PR #1573, opened 2025-02-13
feat(EsZipCodeForm): add contextMessage prop to display optional message above zip input field

Sonnyywithanyextray:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-2115

### ❓ Type of change

- [] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

In es-zip-code-form, I've added new contextMessage prop in the props definition, and added template logic inside the es-form-input component to conditionally display the message. 

In zip-code-form.vue, I added new prop documentation in the propTableRows array.

### 🥼 Testing

Testing via local linking between es-cms-pages, es-ds, changes to the storyblok block schema and modification to test11 in articles/solar/test11 on the Page Zip Code Form (Zip Code) Component layer.:  
https://github.com/EnergySage/es-cms-pages
https://github.com/EnergySage/es-ds

general

### 📝 Checklist

  - [ ] navigating by keyboard
- [x] I have included any Storyblok component schema updates.


ericdouglaspratt (changes requested):
Looks good, just two things:
- I see a CI failure with regard to linting, can you fix?
- Can you add an example to the EsZipCodeForm docs page that illustrates a usage of this new prop?
Thanks!

ericdouglaspratt (changes requested):
I see the CI looks good now, so just this one item still outstanding:
- Can you add an example to the EsZipCodeForm docs page that illustrates a usage of this new prop?

es-ds-docs/pages/organisms/zip-code-form.vue
```diff
@@ -261,6 +262,29 @@ const slotTableRows = [
             </es-row>
         </div>
 
+        <div class="mb-500">
+            <h2>context Message Example</h2>
```
ericdouglaspratt, line 266: Can we correct the capitalization to sentence case per our [typography guidelines for headings](https://new.design.energysage.dev/atoms/typography) like "Context message example"?

es-ds-docs/pages/organisms/zip-code-form.vue
```diff
@@ -261,6 +262,29 @@ const slotTableRows = [
...
+                code input.
+            </p>
+            <es-row class="justify-content-center">
+                <es-col
+                    class="d-flex justify-content-center"
+                    sm="10"
+                    md="8">
+                    <es-zip-code-form
+                        input-id="context-message-example"
+                        context-message="Enter your ZIP code to get started"
```
ericdouglaspratt, line 278: Do you see this context message show up on your local dev server? For some reason, I'm not seeing it on mine; I'm not sure why.

Screenshot 2025-02-14 at 2 36 54 PM

ericdouglaspratt (changes requested):

es-ds-docs/pages/organisms/zip-code-form.vue
Sonnyywithanyextray: Is the contextMessage visible on your end now when testing locally?

es-ds-components/components/es-zip-code-form.vue
```diff
@@ -200,5 +208,11 @@ const handleSubmit = () => {
             top: 0.02em;
         }
     }
+    form {
+        align-items: flex-start !important;
+        &:not(:has([data-context-message])) {
+            align-items: flex-end !important;
```
ericdouglaspratt, line 214: We might want a different solution here - if you click the submit button on any of the examples without filling in any zip code (so that the error message displays), the button becomes misaligned relative to the input field.

Screenshot 2025-02-19 at 7 32 06 AM

Also happens on other examples too:

Screenshot 2025-02-19 at 7 31 58 AM

This is a tricky one. Because both the context message and error message both have an arbitrary height (depending on how much text is passed in and whether it wraps) and are enclosed along with the input in a div, and the button is separate and to the right, there's no way I can think of in CSS to guarantee the button is vertically aligned with the input, so we might have to lean on a little bit of JS to help us here.

There might be a simpler solution than what I'm about to propose (maybe run it by some more people on the team), but this is the best I can think of at the moment:
1. Start off with the styles you have here during the server-side render process - which only apply if there's a context message - this is because (a) you can't do step 2 below on the server-side, and (b) there will pretty much never be an error message displayed in the initial state of the form so this alignment will work for the initial render
2. On mount, if there's a context message, determine the height of the context message (might need to modify EsFormInput to give you a ref to that element that you can work with from inside EsZipCodeForm), remove your `align-items: flex-end`, and set the `padding-top` of a `<div>` you put around the EsButton to that same height as the context message (you'll also need to `watch()` the `contextMessage` prop and update that height if the prop changes, as changing the text may change the height of it) - this means the EsButton will be aligned to the top of its "column" except pushed down the exact height of the context message, so it'll vertically align nicely with the input. And this CSS will need to be scoped only to breakpoints where the button and input appear side-by-side (otherwise there will be way too much space between the input and button on mobile), as determined by `stackBreak` (you might be able to insert a JS variable into CSS using `v-bind` like I did in [EsCarousel](https://github.com/EnergySage/es-ds/blob/main/es-ds-components/components/es-carousel.vue#L384))

es-ds-docs/pages/organisms/zip-code-form.vue
ericdouglaspratt: Yes! I can see it now - nice work!

es-ds-components/components/es-zip-code-form.vue
lgeggleston: While I know it would be ideal to continue using the `extraContext` slot that FormInput already provides, given that this is getting so complex to style what do you think about just recreating the message element separately above so that it doesn't have to disturb the form elements current layout functionality at all? (similar to the static privacy policy below)

Sonnyywithanyextray: @ericdouglaspratt Are the button misalignments still appearing on your end? Attached are how it appears locally for me.
Screenshot 2025-02-19 at 3 39 04 PM
Screenshot 2025-02-19 at 3 44 08 PM
Screenshot 2025-02-19 at 3 44 16 PM
Screenshot 2025-02-19 at 3 44 22 PM

es-ds-components/components/es-zip-code-form.vue
```diff
@@ -105,6 +109,11 @@ const handleSubmit = () => {
         }"
         v-bind="$attrs">
         <div class="d-flex flex-column">
+            <div v-if="contextMessage">
+                <p class="mb-25 font-size-75">
```
ericdouglaspratt, line 113: Small thing: since our performance checks regularly report that our pages have an excessive number of DOM elements, that's something I have my eye on in general - it'd save one element without impacting the layout by removing the `<div>` here and putting the `v-if` onto the `<p>`.

ericdouglaspratt (approved):
One small thing to tweak but good to merge after that's done - nice work!

es-ds-components/components/es-zip-code-form.vue
ericdouglaspratt: Yep looks good now!

PR #1590, opened 2025-04-15
fix: popover hover functionality and arrow alignment

ega713:
### ❓ Type of change

### 📚 Description

- Updated `es-popover` so that when it is triggered by `hover`, it will be dismissed (after a 0.25s timeout) when neither the target nor the popover are being hovered over
- Updated docs to add examples of popovers triggered by `hover`
- Updated `arrow` styles for popover so that arrow is correctly aligned with target on xs viewports (previously was aligned all the way left)

### 🥼 Testing

- Tested locally on docs page:
  - Hovering over icon or popover body keeps the popover visible
  - Hovering over icon -> mouseout closes the popover
  - Hovering over popover body -> mouseout closes the popover
  - Clicking on the icon shows the popover, and the popover stays open until a click outside
  - Tab through to focus on the icon opens the popover until a click outside
- Also tested that arrow points to the correct target across viewports

- I tested this with `hover focus` and `hover click` triggers and got the expected results, but are there other triggers to worry about that might not play well with the hover?
- Any way around having to set the popover hover listener on every `show`?
- Are there best practices for choosing a timeout value? I just picked an arbitrary number to ensure that the popover stays open when the hover is moved from the target to the popover body
- Any other possible simplifications, since tracking whether either the popover body or target are being hovered over added complexity

### 📝 Checklist

  - [ ] navigating by keyboard

### DIFF
--- a/es-ds-components/components/es-popover.vue
+++ b/es-ds-components/components/es-popover.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import type { OverlayPanelState } from 'primevue/overlaypanel';
 import OverlayPanel from 'primevue/overlaypanel';
 
 const props = defineProps({
@@ -35,7 +36,7 @@ const showPanel = (event: Event) => {
     if (event.type !== 'mouseover') {
         triggeredBy.value = event.type;
     } else {
-        const overlayVisible = (op.value as any)?.visible as boolean | undefined;
+        const overlayVisible = (op.value as unknown as OverlayPanelState)?.visible;
         if (!overlayVisible) {
             triggeredBy.value = event.type;
         }
--- a/es-ds-components/components/es-popover.vue
+++ b/es-ds-components/components/es-popover.vue
@@ -27,13 +27,34 @@ const props = defineProps({
 const op = ref<InstanceType<typeof OverlayPanel> | null>(null);
 const slots = useSlots();
 const hasTitle = computed(() => !!slots.title);
+const triggeredBy = ref('');
 const closePanel = () => {
     op.value?.hide();
 };
 const showPanel = (event: Event) => {
+    if (event.type !== 'mouseover') {
+        triggeredBy.value = event.type;
+    } else {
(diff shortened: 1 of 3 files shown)


hroth1994:
👀

es-ds-styles/scss/_popover.scss
```diff
@@ -76,6 +77,10 @@
       bottom: -(variables.$popover-arrow-height) - variables.$popover-border-width + 0.1rem;
       left: 0;
       clip-path: polygon(0% 0%, 50% 100%, 100% 0%);
+ 
+      @include breakpoints.media-breakpoint-down("xs") {
+          left: calc(var(--overlayArrowLeft) - 0.75rem);
```
hroth1994, line 82: nit: extra tab

es-ds-components/components/es-popover.vue
```diff
@@ -27,20 +27,51 @@ const props = defineProps({
...
+    if (!props.triggers.split(' ').includes('hover')) return;
 
+    const targetElement = document.getElementById(props.target);
+    const overlayElement = (op.value as any)?.container as HTMLElement | undefined;
+    overlayElement?.addEventListener('mouseleave', () => {
+        setTimeout(() => {
+            if (targetElement && !targetElement.matches(':hover') && triggeredBy.value === 'mouseover') {
+                closePanel();
+            }
+        }, 250);
```
hroth1994, line 55: Did a little digging, and 2.5ms seems like a good hang time. Can you just add a comment indicating that's what this is doing?

hroth1994 (changes requested):
We may be moving away from this component, but here is some small bits of feedback in the meantime:

hroth1994 (approved):
LGTM! FYI this component is likely to change soon (see https://energysage.atlassian.net/browse/CED-1893)

ega713:
Closing due to component rebuild in https://github.com/EnergySage/es-ds/pull/1592. Temporary local changes have been made in https://github.com/EnergySage/quote-frontend/pull/469 to handle these cases.

PR #1591, opened 2025-04-15
fix: update EsCollapse to use aria-labbeledby for accessibility

Sonnyywithanyextray:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-2203?atlOrigin=eyJpIjoiMWI5Njg5ZTkyNzM5NDAzYzk5NmZiN2E2NzYyODMwOTkiLCJwIjoiaiJ9

### ❓ Type of change

### 📚 Description

Previously, screen readers announced an incorrect header for the EsCollapse toggle. This change updates EsCollapse to use aria-labelledby on the header slot, so that the correct text is announced.

### 🥼 Testing

Tested locally at http://localhost:8500/molecules/collapse. The tab key focuses the toggle button and pressing Enter expands and collapses the panel. I also used VoiceOver on macOS so the toggle button now announces the correct header text through aria-labelledby. Images of using tab key to focus the toggle button and pressing enter, as well as using VoiceOver, is attached.

Screenshot 2025-04-15 at 4 10 24 PM
Screenshot 2025-04-15 at 4 10 57 PM

general

### 📝 Checklist

  - [ ] navigating by keyboard

### DIFF
--- a/es-ds-components/components/es-collapse.vue
+++ b/es-ds-components/components/es-collapse.vue
@@ -53,7 +53,7 @@ const onClick = ({ value }: { value: boolean }) => {
     <panel
         :collapsed="!expanded"
         :header="''"
-        :toggle-button-props="{ 'aria-labelledby': 'collapse-header'}"
+        :toggle-button-props="{ 'aria-labelledby': 'collapse-header' }"
         :pt="{
             header: 'd-flex flex-column justify-content-center position-relative py-100',
             icons: 'h-100 position-absolute w-100',
--- a/es-ds-components/components/es-collapse.vue
+++ b/es-ds-components/components/es-collapse.vue
@@ -52,7 +52,8 @@ const onClick = ({ value }: { value: boolean }) => {
 <template>
     <panel
         :collapsed="!expanded"
-        :header="expanded ? 'click to collapse content' : 'click to expand content'"
+        :header="''"
+        :toggle-button-props="{ 'aria-labelledby': 'collapse-header'}"
         :pt="{
             header: 'd-flex flex-column justify-content-center position-relative py-100',
             icons: 'h-100 position-absolute w-100',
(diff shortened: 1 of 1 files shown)


nathanielwarner:
👀

Sonnyywithanyextray:
Just realized I requested more then 3 reviewers on this. My mistake.

es-ds-components/components/es-collapse.vue
```diff
@@ -74,7 +75,9 @@ const onClick = ({ value }: { value: boolean }) => {
         toggleable
         @toggle="onClick">
         <template #header>
-            <slot name="title" />
+            <div id="collapse-header">
```
nathanielwarner, line 78: Clever solution here!

es-ds-components/components/es-collapse.vue
```diff
@@ -52,7 +52,8 @@ const onClick = ({ value }: { value: boolean }) => {
 <template>
     <panel
         :collapsed="!expanded"
-        :header="expanded ? 'click to collapse content' : 'click to expand content'"
+        :header="''"
```
nathanielwarner, line 55: Can we get rid of the `header` prop altogether? Passing a blank string is causing `aria-label=""` to be placed on the header which we don't want, even though it doesn't seem to cause an issue with VoiceOver.

nathanielwarner (changes requested):

es-ds-components/components/es-collapse.vue
nathanielwarner: https://nuxt.com/blog/v3-10#ssr-safe-accessible-unique-id-creation

es-ds-components/components/es-collapse.vue
```diff
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import Panel from 'primevue/panel';
+import { useId } from '#imports';
```
hroth1994, line 3: quick nit - `useId` is available in Nuxt without needing to import it

nathanielwarner (approved):
Nice work!

PR #1593, opened 2025-04-17
fix: es-tabs when comments are children  

susri523:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CORE-6387

### ❓ Type of change

### 📚 Description

- filter out the comments from the initialChildren that are extracted from the slot 

### 🥼 Testing

- Tested locally on docs page:
  - confirmed existing functionality did not change
  - checked that passing an es-tab with a v-if="false" can still render 
  - checked that adding in comments into the slots still allows the component to render  

used the following in the docs page to test on main branch vs. this one 
```
            <es-tabs>
                <es-tab title="Item one">
                    <p>Content one</p>
                </es-tab>
                <es-tab title="Item two">
                    <p>Content two</p>
                </es-tab>
                <es-tab title="Item three">
                    <p>Content three</p>
                </es-tab>
                <!--  -->
                <es-tab 
                    v-if="false"
                    title="Item four">
                    <p>Content four</p>
                </es-tab>
            </es-tabs>
```

- Any better/cleaner ways to filter out the unnecessary children here? 

### 📝 Checklist

  - [ ] navigating by keyboard

### DIFF
--- a/es-ds-components/components/es-tabs.vue
+++ b/es-ds-components/components/es-tabs.vue
@@ -1,12 +1,15 @@
 <script setup lang="ts">
 import TabPanel from 'primevue/tabpanel';
 import TabView from 'primevue/tabview';
+import { Comment } from 'vue';
 
 // allow use of v-model on this component
 const model = defineModel<number>();
 
-// get the list of elements provided as children to the default slot
-const initialChildren = useSlots().default?.() || [];
+// get the list of elements provided as children to the default slot - filtering out comments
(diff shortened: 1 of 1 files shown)


ericdouglaspratt (approved):
Tested locally, looks good!

Good catch that code comments break rendering of EsTabs. We may eventually switch away from PrimeVue for components like this and the accordion that have such a direct-children requirement (which IMO is an unnecessarily restrictive and fragile implementation), but this definitely helps fix issues with the current one for now.

PR #1598, opened 2025-04-21
fix: Add to setup instructions

lgeggleston:
### 🔗 Linked issue

N/A

### ❓ Type of change

### 📚 Description

- Added the lines we have elsewhere that indicate the prerequisites for installing with npm in this repo.  Ran into this because Laura was trying to set up es-ds recently.

-

### 📝 Checklist

  - [ ] navigating by keyboard

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,10 +60,12 @@ Bootstrap Vue styles present in `es-ds-styles`. Those are deprecated and will be
 in a future version.
 
 **es-ds-components**
+
 - [es-ds-components](./es-ds-components/) contains Vue 3 components for use in Nuxt 3 projects.
 It is primarily based on [PrimeVue](https://primevue.org/) and [Reka UI](https://reka-ui.com/), with some customizations specific to EnergySage.
 
 **es-ds-docs**
+
 - This is the design system documentation site powered by Nuxt 3.
 - It also serves a reference for
 how to integrate `es-ds` packages into a Nuxt 3 project.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,7 +87,14 @@ origin  git@github.com:EnergySage/es-ds.git (push)
 
 ### Installing Dependencies and Linking packages
 
-1. `make install && make symlink` - installs all packages from npm
+1. You will need access to the Consumer-Education-Dev AWS profile, and to clone <https://github.com/EnergySage/es-dev-utils> to run the below scripts.
```
lauramichellecp, line 90: I've been using the installer dev account so far, so I don't think people will need access to the consumer ed dev for local development. The account is just needed for the codeartifact piece, and you guys probably dont want people outside your team added to your aws accounts.

README.md
```diff
@@ -85,7 +87,14 @@ origin  git@github.com:EnergySage/es-ds.git (push)
...
+
+```
+source ../es-dev-utils/aws_login.sh Consumer-Education-Dev
+../es-dev-utils/npm_config.sh Consumer-Education-Dev
```
lauramichellecp, line 94: This is coming from someone with little node knowledge here, so maybe this is obvious.

While I was trying to set this locally, I figured out there were a couple of places where the configuration script needed to run. Specifically, every directory with a `package.json` in it, since it generates npmrc files for each. I had to run it in /es-ds-styles and /es-ds-components (maybe also /es-ds-docs, not sure) before being able to run the next command without encountering errors. Might be worth noting that.

ericdouglaspratt: There are only public NPM packages in this repo - there should be no need to run the authentication script at all, nor do I think we should be exposing our security setup steps publicly here (despite the es-dev-utils repo in question being private).

lauramichellecp: Oh, interesting. If the question is whether this should be documented or not, I'm not sure. But at least for me I wasn't able to run anything like `make install` or symlink without running these first because of this error:
```
npm error code E401
npm error Unable to authenticate, your authentication token seems to be invalid.
npm error To correct this please try logging in again with:
npm error  npm login
```

My understanding was that we needed to hook this up to [codeartifact](https://docs.aws.amazon.com/codeartifact/latest/ug/npm-auth.html#configure-npm-login-command)

Unless I've set this up wrong locally, and im not supposed to be using code artifact

tomleo: So some AWS commands configure codeartifact credentials at the user level instead of the repo level. You likely have an `.npmrc` file in your home folder. Removing that file should allow you to install directly from the public npm.

PR #1600, opened 2025-04-23
feat: simplify EsPopover layout and allow text to go full width

ericdouglaspratt:
### 🔗 Linked issue

- n/a

### ❓ Type of change

### 📚 Description

- All previous iterations of EsPopover had hardcoded right-side padding across the entire height of the popover contents in order to make room for the close button - this PR floats the close button to the right so text and other content can go full width beneath it and better make use of all available space (especially helpful for when the popover contains a lot of text)
- Because of this, the `cta` slot is no longer needed to allow buttons to go full width, so it's kept for backwards compatibilty but silently deprecated and no longer documented

### Before (side padding)
Screenshot 2025-04-23 at 7 40 56 AM

### After (text wrapping around the floated close button)
Screenshot 2025-04-23 at 7 49 34 AM

### 🥼 Testing

- Tested locally on EsPopover and EsTooltip docs pages
- Tested by npm linking to Pages and viewing solar data explorers page

- Overall

### 📝 Checklist

  - [ ] navigating by keyboard

### DIFF
--- a/es-ds-components/components/es-popover.vue
+++ b/es-ds-components/components/es-popover.vue
@@ -42,7 +42,7 @@ const emit = defineEmits(['update:show']);
                 :collision-padding="collisionPadding"
                 :side="side">
                 <popover-close
-                    class="es-popover-close align-items-center d-flex justify-content-center float-right mb-50 ml-50"
+                    class="es-popover-close align-items-center d-flex justify-content-center float-right mb-50 ml-50 position-relative"
                     :class="{ 'es-popover-close--light': variant === 'light' }">
                     <span class="sr-only">Close</span>
                     <icon-x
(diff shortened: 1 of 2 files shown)


mpleroux (commented):
Everything works as you said. I tested it in Chrome, Firefox, and Safari.

One minor quirk in Safari. When the popup first opens the close button appears with a blue rectangle around it. Clicking off the element works as expected.

It looks like the hover or focus states, but I'm having trouble examining it in the browser dev tools.

ericdouglaspratt:
@mpleroux Yeah that was pre-existing before this PR - upon open of the popover, it becomes a "focus trap" and focus is first given to the close button (since it's the first focusable thing inside it) - so the outline is accurate, as the close button does have focus, but we may be able to switch up the styles so it triggers on `:focus-visible` instead of `:focus` so sighted users aren't unnecessarily distracted by it. However, I'd consider that a separate fix from this PR, since it was happening prior.

mpleroux (approved):

PR #1602, opened 2025-04-23
feat: Add closeable and showFooterSeparator props to es-modal

lauramichellecp:
### ❓ Type of change

- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

Following the modal component in this [design](https://www.figma.com/design/oL5VjPJahIa6lC9wfOknZq/Billing-screens?node-id=0-1&p=f&t=TrqcN2ZRPGrlvlv2-0),
- Added the `closable` prop to es-modal to hide/show the close button, default set to true
- Added the `showFooterSeparator` prop to es-modal to hide the line separating the content from the modal footer, default set to true
- Used the modal-title slot to determine whether to show the modal header or not
- Added an example of the modal without a header or the footer separator in /es-ds-docs

### 🥼 Testing

- Tested by comparing the different modals in es-ds-docs locally vs. in https://new.design.energysage.dev/ and ensured there were no regressions introduced with the changes.

- This seemed like the only way to conditionally add or remove the header rather than styling it to not display on the page. Hoping this is the right type of contribution
- Any instructions on verifying accessibility? 

### 📝 Checklist

  - [ ] navigating by keyboard


ItsEricSun:
Modal LGTM!
image

es-ds-components/components/es-modal.vue
```diff
@@ -10,6 +10,10 @@ const props = defineProps({
         type: Boolean,
         default: false,
     },
+    hideHeader: {
+        type: Boolean,
+        default: false,
+    },
```
ericdouglaspratt, line 16: Rather than adding a new prop for this, it might be simpler to check whether they've passed in that `<template #modal-title> Arcadia Reviews </template>` slot or not.

You can do `const slots = useSlots();` in here to get access to the slots, and then doing a boolean check on `slots['modal-title']` should tell you whether or not the slot was provided.

es-ds-components/components/es-modal.vue
```diff
@@ -27,7 +31,7 @@ const modalPt = {
         class: 'modal-dialog modal-dialog-scrollable modal-content',
     },
     header: {
-        class: 'modal-header',
+        class: props.hideHeader ? '' : 'modal-header',
```
ericdouglaspratt, line 34: This could be `class: slots['modal-title'] ? 'modal-header' : '',`

es-ds-components/components/es-modal.vue
```diff
@@ -75,9 +79,12 @@ const getSizeClass = computed(() => {
         modal
         :class="getSizeClass"
         :pt="modalPt"
+        :closable="!hideHeader"
```
ericdouglaspratt, line 82: I'd recommend controlling this separately from the header, by exposing a new boolean `closeable` prop, defaulted to true, and passing it in here.

es-ds-components/components/es-modal.vue
```diff
@@ -75,9 +79,12 @@ const getSizeClass = computed(() => {
...
         dismissable-mask
         @update:visible="onChange">
-        <template #header>
+        <template
+            v-if="!hideHeader"
+            #header>
```
ericdouglaspratt, line 87: This could be:

```
<template
    v-if="slots['modal-title']"
    #header>
```

es-ds-docs/pages/molecules/modal.vue
```diff
@@ -106,6 +112,42 @@ onMounted(async () => {
...
                 </template>
             </es-modal>
+
+            <es-modal
+                id="no-header-modal"
+                :visible="noHeaderModalVisible"
+                hide-header
+                size="lg"
+                @hidden="noHeaderModalVisible = false">
+                <template #modal-title> Arcadia Reviews </template>
```
ericdouglaspratt, line 122: And then once the above setup is complete, removing this should remove the header.

ericdouglaspratt (changes requested):
Looks great but I'd recommend a few quick simplifications and one semantic separation.

Also, per your linked design, is it worth having this example also demonstrate how to control the inner padding of EsModal by passing in `class="px-200 py-150"` to it?

lauramichellecp:
> Looks great but I'd recommend a few quick simplifications and one semantic separation.
> 
> Also, per your linked design, is it worth having this example also demonstrate how to control the inner padding of EsModal by passing in `class="px-200 py-150"` to it?
> 
> Also also, it looks like your design doesn't have that big line separator between the buttons and the modal content - worth tweaking that here? Could also be done separately later.

Thanks for the review, this is really helpful! I'll make the tweaks. 

On the big line separator... I'm still unsure what the line is when it comes to styling things in the design system components vs. tweaking existing components in our microfrontend. If I were to tweak change this here, are you suggesting conditionally removing it as well by exposing another boolean?

lauramichellecp:
Thanks again for the review! I've made all the suggested changes. 

My final question is about the padding on the top of the footer. Is that something that makes sense to change here as well, or not?

es-ds-styles/scss/_modal.scss
```diff
@@ -112,13 +112,15 @@
...
 
-  &::before {
-    border-top: variables.$modal-footer-border-width solid variables.$modal-footer-border-color;
-    content: '';
-    left: 0;
-    position: absolute;
-    right: 0;
-    top: variables.$modal-footer-padding-bottom;
+  &.footer-separator {
+    &::before {
```
ericdouglaspratt, line 116: Yeah I think it's fair to move the `padding-top: variables.$modal-footer-padding-top;` above into the `&.footer-separator` class, and have the default up top be `padding-top: 1rem`.

ericdouglaspratt (changes requested):
Looks great! Just one thing to answer your question.

ericdouglaspratt (approved):
Nice work!

PR #1618, opened 2025-06-06
fix: upgrade to nuxt 3 17

nathanielwarner:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-2358

### ❓ Type of change

### 📚 Description

- Upgraded to Nuxt 3.17, alongside other dependency upgrades that happen as a result of clearing out `node_modules` and `package-lock.json`.
- Wrap the Reka UI based components (Tooltip and Popover) in ClientOnly when in the dev server, which prevents the SSR issue.
  - Other ideas from [Eric's troubleshooting doc](https://docs.google.com/document/d/1nmgm4RG9s1UsGF0ra5yV-uGWJvqHXsfdcoQOQdivPr8/edit?tab=t.0) were attempted without success. Also tried aliasing instances of Vue modules using Vite, as well as forcing deduplication. Using Yalc instead of npm link could be an alternative (and seems to be what people do with React due to analogous issues), but I think it's a lot simpler to just have this ClientOnly workaround since only two components are affected. Using Yalc would mean that hot reloading is no longer possible, which would be really painful for design system local dev.

### 🥼 Testing

Tested in dev server and static generation. Note that in the dev server, the popover and tooltip components no longer render until the time of hydration, while they are still rendered during static generation as before, which prevents a layout shift on our production site. Also checked each component docs page for errors (including hydration errors) during SSR, CSR, and SSG. There is one pre-existing hydration error on the data table docs page (ticket pending, I think the best thing would be to switch to the custom implementation I made a few months ago for the CMS repos...)

- General

### 📝 Checklist

  - [x] navigating by keyboard

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 3.1.6 - 2025-06-06
+
+- Upgraded dependencies, including Nuxt (3.15 to 3.17)
+
 ## 3.1.5 - 2025-05-13
 
 - Added `sideImage` slot, `sideImageContainerClass` prop, and `semi-wide` variant to EsCtaBanner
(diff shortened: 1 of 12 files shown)


ericdouglaspratt (approved):
Fantastic solution! Tested locally on dev and generate, works great

Let's goooooooo

PR #1622, opened 2025-07-07
feat: match EsProgress to design

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CED-2478

### ❓ Type of change

### 📚 Description

- Brings EsProgress into alignment with the updated Figma designs that have been out for many months
- Added `formatter`, `showCircle`, and `valueClass` props
- This is a breaking change as all instances of EsProgress, especially those with custom styles applied, will need to be reviewed and adjusted for visual consistency (likely those custom styles can be removed, in cases where they were applied to match the design)

screencapture-localhost-8500-molecules-progress-2025-07-07-18_11_08

### 🥼 Testing

- Tested locally - and did a quick [code search](https://github.com/search?q=org%3AEnergySage%20es-progress&type=code) to see how it's being used

- Overall

### 📝 Checklist

  - [ ] navigating by keyboard

### DIFF
--- a/es-ds-docs/pages/molecules/progress.vue
+++ b/es-ds-docs/pages/molecules/progress.vue
@@ -5,12 +5,15 @@ const compCode = ref('');
 const docCode = ref('');
 
 const propTableRows = [
-    ['value', 'Number', '20', 'Required. A value between 0 and 100 representing the progress'],
-    ['height', 'String', '0.3125rem', 'Specifies height of the progress bar'],
-    ['showValue', 'Boolean', 'false', 'When present, it allows the progress percentage visible on the progress bar'],
+    ['formatter', 'Function', '(val: number) => `${val}%`', 'Allows controlling the display of the value displayed when showValue is set to true.'],
+    ['height', 'String', '0.125rem', 'Specifies height of the progress bar'],
+    ['showCircle', 'Boolean', 'true', 'Set to false to hide the indicator circle.'],
+    ['showValue', 'Boolean', 'false', 'When present, it shows the progress bar value below the bar.'],
+    ['value', 'Number', 'n/a', 'Required. A value between 0 and 100 representing the progress'],
+    ['valueClass', 'String', "''", 'Allows control of styling for the value displayed when showValue is set to true.'],
 ];
 const propTableWidths = {
-    md: ['2', '2', '2', '6'],
+    md: ['2', '2', '3', '5'],
 };
 
 onMounted(async () => {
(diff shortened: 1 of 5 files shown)


es-ds-docs/pages/molecules/progress.vue
```diff
@@ -5,12 +5,20 @@ const compCode = ref('');
...
-    ['value', 'Number', '20', 'Required. A value between 0 and 100 representing the progress'],
-    ['height', 'String', '0.3125rem', 'Specifies height of the progress bar'],
-    ['showValue', 'Boolean', 'false', 'When present, it allows the progress percentage visible on the progress bar'],
+    [
+        'formatter',
+        'Function',
+        '(val: number) => `${val}%`',
+        'Allows controlling the display of the value displayed when showValue is set to true.',
+    ],
+    ['height', 'String', '0.125rem', 'Specifies height of the progress bar'],
```
hroth1994, line 14: Some of the prop descriptions end with periods and some don't. If that's unintentional I would remove the trailing periods.

es-ds-docs/pages/molecules/progress.vue
```diff
@@ -29,54 +37,161 @@ onMounted(async () => {
...
+        <div class="my-500">
+            <h2>Animated transition</h2>
+            <p>
+                If the value of the progress bar is changed programmatically, it will animate to the new value. Try
+                clicking one of the buttons below to add or subtract from the progress bar's value.
+            </p>
+
+            <es-progress
+                class="mb-100"
+                :value="value" />
```
hroth1994, line 74: This value is affected by the buttons in the next example too. I would consider make a new variable for this example.

es-ds-docs/pages/molecules/progress.vue
```diff
@@ -29,54 +37,161 @@ onMounted(async () => {
...
+            <es-progress
+                class="mb-200"
+                height="0.375rem"
+                :show-circle="false"
+                :value="100" />
         </div>
-        <p class="mb-300">value: {{ value }}</p>
+
         <div class="mb-500">
             <h2>EsProgress props</h2>
```
hroth1994, line 196: Since this will changes to es-progress props in a future ticket, I would change it while you're already in here

hroth1994 (approved):
Nice job! Just a few nits that could be taken or left.

es-ds-docs/pages/molecules/progress.vue
ericdouglaspratt: I'm honestly still not sure if I agree with that change - short-term, this could just be "Props" rather than having to repeat the name of the component for every table (which we're not doing for the headers for the examples above).

Long-term, I want a site-wide toggle for "dev mode" that reveals syntax-highlighted code snippets for each example, and perhaps a separate tab for eng-specific things like the list of props and slots. This is so that designers and other non-technical stakeholders viewing the design system can find a friendly, readable, accessible page without it by default being cluterred by technical nonsense - "es-progress" vs. "EsProgress" means nothing to them, especially since the nav bar on the left says "Progress".

es-ds-docs/pages/molecules/progress.vue
ericdouglaspratt: Since one of the descriptions contains two sentences (which would require a period to separate them), I opted for adding periods to all of them.

PR #1638, opened 2025-07-23
fix: various build and release issues

nathanielwarner:
- Exempt changelogs from prettier
- Include CI in the release process

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,15 +1,9 @@
 name: ci
 
 on:
-    push:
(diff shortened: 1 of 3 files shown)


ericdouglaspratt (approved):
Seems fair that CI should continue to run on PRs, and it is true that in the past, CI has run on main after a PR merge, at which point a failure is too late. We do require PR branches to be up to date before merging, so this should catch every change.

I'm assuming that release-please will do something with the new CI job that's now in release-please.yml?

PR #1642, opened 2025-07-25
feat: CED-2491 create es-dropdown

martinsieving:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/CED-2491

### ❓ Type of change

### 📚 Description

- Create a new dropdown component, following the [design](https://www.figma.com/design/G3Zq89yg3vek9b4JrnlSYn/ESDS-2.0?node-id=16868-14045&m=dev)
- Add a docs page for the component

### 🥼 Testing

- Tested locally.
- Dropdown always appears below the input box.

- Accessibility testing
- Dropdown window is visually merged with the input box in a sorta hacky way, and does not work if the dropdown appears above the input (which is why we are forcing it to be below)
  - This is causing some inconsistencies with boarder color for different focus states.
- Currently have it set up so we can pass in a list of objects and tell it what values to use for the label, as well as what value to return.
  - This wasn't part of the design, so it should maybe get taken out.
- Just in general if stuff has been made with customization in mind. If it isn't easily customizable it should probably get reworked.

### 📝 Checklist

  - [ ] navigating by keyboard


es-ds-styles/scss/_dropdown.scss
```diff
@@ -201,3 +201,137 @@
   padding: variables.$dropdown-item-padding-y variables.$dropdown-item-padding-x;
   color: variables.$dropdown-link-color;
 }
+
+// ==========================================
+// ES-DROPDOWN CUSTOM STYLES
+// ==========================================
+
+// Custom es-dropdown input styling
+.es-dropdown-input {
```
ericdouglaspratt, line 210: Can we throw a `user-select: none;` in here so that clicking a bunch of times to open and close the dropdown only opens and closes the dropdown and doesn't highlight all the text within it?

https://github.com/user-attachments/assets/e8fc124f-c520-49cc-91da-54556e132ad0

es-ds-styles/scss/_dropdown.scss
```diff
@@ -201,3 +201,137 @@
...
+  justify-content: space-between;
+  padding: 16px 16px 16px 32px;
+  font-family: variables.$input-font-family;
+  font-size: variables.$input-font-size;
+  font-weight: variables.$input-font-weight;
+  line-height: variables.$input-line-height;
+  text-decoration: none;
+  transition: border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out;
+  white-space: nowrap;
+  width: 325px;
```
ericdouglaspratt, line 228: Our dropdowns are definitely going to need to vary in width based on the use case and the text inside them, so I don't think we should hardcode a width here. Dropdowns should just always take up the full width of their container and we can let the downstream app authors limit the width as necessary for their use case.

es-ds-components/components/es-dropdown.vue
```diff
@@ -0,0 +1,82 @@
...
+            {{ props.title }}
+        </label>
+        <label
+            v-else
+            :for="id"
+            class="sr-only">
+            Select an option
+        </label>
+        <dropdown
+            :id="id"
```
ericdouglaspratt, line 41: I noticed that clicking the label doesn't automatically give focus to the field, which it should. Inspecting the DOM shows that this id is applied to the `<div>` wrapper rather than the `<span tabindex="0">` that is focusable. Per the [PrimeVue Dropdown API Docs](https://v3.primevue.org/dropdown/), it looks like this should be `inputId` instead of `id` - that might solve it?

es-ds-styles/scss/_dropdown.scss
```diff
@@ -201,3 +201,137 @@
...
+  transition: background-color 0.15s ease-in-out;
+  width: 100%;
+
+  &:hover {
+    background-color: variables.$blue-50;
+  }
+
+  &:active {
+    background-color: variables.$blue-100;
+    font-weight: variables.$font-weight-bold;
```
ericdouglaspratt, line 316: If design agrees to change the focus state of the overall dropdown away from the bold and towards an outline, can you check with them to maybe revisit the bolding on "pressed" state as well? Because of the animation, the background color fades in more slowly than the font bolding (which takes effect immediately), meaning there are two pieces of the "pressed" state that arrive at different times, which looks odd to me. I might recommend to design that the background color change is enough.


https://github.com/user-attachments/assets/666f6406-b8b5-4541-98c4-6e55528876e1

es-ds-styles/scss/_dropdown.scss
```diff
@@ -201,3 +201,137 @@
...
+    border-bottom-left-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+}
+
+// Dropdown panel styling
+// Top corners have no radius so that it visually merges with the input
+.es-dropdown-panel {
+  background-color: variables.$white;
+  border: 1px solid variables.$gray-500;
```
ericdouglaspratt, line 274: Nit: can we check with design about whether we want to provide any elevation to the dropdown's popup menu, in case of visual collision with another similar-looking input or dropdown, which, depending on how they're positioned, could cause confusion about where this dropdown ends and the other one begins?

Screenshot 2025-07-30 at 1 41 41 PM

One option could be to reuse the subtle shadow from the light variant of EsPopover.

Screenshot 2025-07-30 at 1 43 35 PM

es-ds-docs/pages/molecules/dropdown.vue
```diff
@@ -0,0 +1,138 @@
...
+
+const selectedFruit = ref<string | undefined>(undefined);
+const selectedColor = ref<string | undefined>(undefined);
+const selectedSize = ref<string | undefined>(undefined);
+
+const fruits = ['Apple', 'Banana', 'Orange', 'Grape', 'Strawberry'];
+
+const colors = ['Red', 'Orange', 'Yellow', 'Green', 'Blue', 'Indigo', 'Violet'];
+
+const sizes = ['XS', 'S', 'M', 'L', 'XL', 'XXL'];
```
ericdouglaspratt, line 24: Can we have an example on this page that showcases passing in an array of objects with "option" and "value" keys, like [this example from PrimeVue](https://v3.primevue.org/dropdown/#template)? Our immediate use cases will very likely involve options having user-facing text with a different string as the value we save behind the scenes.

es-ds-components/components/es-dropdown.vue
```diff
@@ -0,0 +1,188 @@
...
+            :placeholder="placeholder"
+            :pt="{
+                panel: { class: 'es-dropdown-panel bg-white rounded-xs' },
+                wrapper: { class: 'es-dropdown-wrapper' },
+                list: { class: 'p-0 m-0 list-unstyled' },
+                item: {
+                    class: [
+                        'es-dropdown-item d-flex justify-content-between p-100 pl-200',
+                        {
+                            'input-focused': isFocused,
```
ericdouglaspratt, line 68: Something about this logic is causing this to happen: when you use the mouse to click on the dropdown, opening the menu, then Command-Tab to another window, then Command-Tab back, the item has a focus outline when it shouldn't.

https://github.com/user-attachments/assets/6be598ce-98a7-4f35-93fb-db49ed91fb66

As a general rule, I don't think the user should ever see these ugly focus outlines unless they're tabbing or arrowing  through via keyboard.

es-ds-components/components/es-dropdown.vue
```diff
@@ -0,0 +1,188 @@
...
+.es-dropdown-input {
+    background-clip: padding-box;
+    border: variables.$border-width solid variables.$gray-900;
+    height: variables.$input-btn-height;
+    transition: variables.$input-transition;
+    user-select: none;
+
+    &:hover,
+    &:active {
+        border-color: variables.$blue-300;
```
ericdouglaspratt, line 109: The text input's border becomes blue when it receives focus and stays blue throughout your interaction with it, until you're done and move on to another input.

https://github.com/user-attachments/assets/0cd1e481-b629-4bac-a8f3-18a6ca3d4451

Whereas the dropdown border currently only becomes blue on hover and goes back to being gray as soon as you move your mouse elsewhere or to the menu. I'd expect it to match the behavior of input, which is that the border becomes blue and stays blue during the entire duration of the time the user is engaging with the dropdown, meaning focused or menu open.

https://github.com/user-attachments/assets/79704436-b658-46b6-9121-f24983839411

es-ds-components/components/es-dropdown.vue
```diff
@@ -0,0 +1,188 @@
...
+            Select an option
+        </label>
+        <dropdown
+            :input-id="id"
+            :class="[
+                'es-dropdown-input d-flex align-items-center bg-white rounded-xs justify-content-between p-100',
+                {
+                    disabled: disabled,
+                    'es-dropdown-open': isOpen,
+                    focused: isFocused && !isOpen,
```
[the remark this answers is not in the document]
hroth1994: Yep I noticed this during the demo, I updated the logic and tested it for all cases of click/tab to open and believe the focus states are solid now.

es-ds-components/components/es-dropdown.vue
```diff
@@ -0,0 +1,198 @@
+<script setup lang="ts">
+import Dropdown from 'primevue/dropdown';
```
ericdouglaspratt, line 2: What do you think about naming our component a "dropdown select" vs. just "dropdown"? I can see a future - once we redo the nav and make smaller support "dropdown menu" and "mega menu" type components - where the word "dropdown" could get a bit overloaded in that world (which is perhaps why PrimeVue v4 renames it to "Select", which works but IMO "dropdown select" is even clearer and catches both words in case people search for one or the other).

Here's what Google AI had to say about it:

> The terms "dropdown," "select," and "dropdown select" are often used interchangeably in user interface design, but they can have slightly different connotations or be used to emphasize specific aspects of the component.
> 
> Dropdown:
> This is a general term referring to any UI element that, when clicked or hovered over, reveals a list of options or commands that "drop down" from the main element. It can encompass various functionalities, including navigation menus, lists of actions, or even form inputs.
>
> Select:
> This term specifically refers to the HTML `<select>` element, which is a form control used to present a list of options from which a user can choose one or more. It is primarily used for data input within forms, such as selecting a country or a product category.
> 
> Dropdown Select:
> This term combines "dropdown" and "select" to explicitly describe a form input element that functions as a dropdown list for selecting an option. It emphasizes both the visual behavior (dropping down) and the functional purpose (selecting data for a form).
> 
> In essence, while all "select" elements are typically "dropdowns" in their visual behavior, not all "dropdowns" are "select" elements in the technical sense of being a form input. "Dropdown select" clarifies that the component is a dropdown specifically used for selection within a form context.

es-ds-components/components/es-dropdown.vue
```diff
@@ -0,0 +1,198 @@
...
+        cursor: not-allowed;
+        span[data-pc-section='input'] {
+            color: variables.$gray-500;
+        }
+    }
+}
+
+.es-dropdown-panel {
+    border: variables.$border-width solid variables.$gray-900;
+    box-shadow: variables.$popover-box-shadow;
```
ericdouglaspratt, line 152: Have we run the shadow and overall most recent visual treatment by Asheeta?

I think the shadow does help in this case to distinguish the panel from other things on the page below, but IMO it'd be even nicer to have the whole thing have a shadow when open so it looks like one thing instead of two, but since they're completely separate DOM elements, it may be difficult to do that without one's shadow being over the other.

Screenshot 2025-08-07 at 7 20 40 AM

I also wish we could more visually join the dropdown panel with the dropdown input itself (with the corner rounding and a single 1px border instead of two 1px borders, depending on whether they end up the same color or not), but since we don't know whether it's above or below, there probably isn't an easy way of doing that without grabbing refs to each piece and directly observing the `offsetTop` and which one is higher on the page (which is overly complicated and might end up having a delay anyway).

The Figma design also didn't account for the fact that the border would be blue when the dropdown is active, but the panel border would be dark gray, meaning their design for a tight joining between the two wouldn't even really work, unless we forgo the blue border (which then would behave differently from inputs).

Screenshot 2025-08-07 at 7 30 31 AM

Screenshot 2025-08-07 at 7 30 01 AM

Overall I'm just curious if Asheeta has any strong feelings about visual tweaks we could make within our constraints to make this look a little better. Maybe the three of us can just sync up today and have a final pass at it?

es-ds-components/components/es-dropdown.vue
```diff
@@ -0,0 +1,198 @@
...
+        border-color: variables.$blue-600;
+        outline: 0.125rem solid variables.$blue-600;
+        outline-offset: 0.125rem;
+        span[data-pc-section='input'] {
+            outline: 0;
+        }
+    }
+
+    span[data-pc-section='input'] {
+        color: variables.$input-color-placeholder;
```
ericdouglaspratt, line 133: I realize this color is what the design has for "Dropdown text", but (a) it's very unclear from the design whether that's only meant for placeholder text or the actual input value, and (b) I suspect they weren't thinking about how actual values would look.

Currently, this makes a selected value look like placeholder text rather than an actual value, and doesn't match how we display input values.

Screenshot 2025-08-07 at 7 32 24 AM

We actually don't need to do `span[data-pc-section='input']` here because the passthrough options give us an `input` item we can pass classes to. If we rename the `.es-dropdown-input` class of the container to `.es-dropdown`, we can do this passthrough config for that:

```
input: {
    class: [
        'es-dropdown-input',
        {
            'es-dropdown-input--placeholder': !modelValue,
        },
    ],
},
```

And then we can do this style below:

```
.es-dropdown-input {
    &:focus-visible {
        outline: 0;
    }

    &--placeholder {
        color: variables.$input-color-placeholder;
    }
}
```

Which gives us the placeholder color when it's placeholder text (and doesn't have a value), and a real color when it's an actual value.

Screenshot 2025-08-07 at 7 55 57 AM

Screenshot 2025-08-07 at 7 56 02 AM

es-ds-docs/pages/examples/form-field-validation.vue
```diff
@@ -30,6 +31,9 @@ const rules = {
             [vuelidateKeys.HAS_UPPERCASE_LETTER]: vuelidateHasUppercaseLetter(1),
             [vuelidateKeys.HAS_LOWERCASE_LETTER]: vuelidateHasLowercaseLetter(1),
         },
+        contactMethod: {
+            [vuelidateKeys.REQUIRED]: vuelidateRequired,
```
ericdouglaspratt, line 35: Also I just realized this form is broken right now because this field is marked as required in the validation config, but (a) the label doesn't have an asterisk next to it denoting that the field is required, and (b) EsDropdown doesn't yet support display of an error message. If you fill out all of the form fields except the dropdown, it doesn't let you submit, but it also doesn't display any error message.

We'll need to support config and display of an error message in a similar way to EsFormInput.

Screenshot 2025-08-07 at 8 05 48 AM

ericdouglaspratt (changes requested):

es-ds-components/components/es-dropdown.vue
hroth1994: Wow nice find!

es-ds-docs/pages/examples/form-field-validation.vue
hroth1994: Added error state

es-ds-components/components/es-dropdown.vue
hroth1994: I ran it by Asheeta and updated the dropdown to a lighter gray to be able to differentiate from other inputs

ericdouglaspratt (changes requested):
Looking pretty good! Just a couple of final cleanup items.

es-ds-components/components/es-dropdown-select.vue
```diff
@@ -0,0 +1,242 @@
...
+    <div
+        class="input-wrapper justify-content-end"
+        :class="$attrs.class">
+        <label
+            :id="labelId"
+            :for="id"
+            :class="{ 'sr-only': label === '' }"
+            @click="isLabelClicked = true">
+            <span v-if="label !== ''">{{ label }}</span>
+            <span v-else>Select an option</span>
```
ericdouglaspratt, line 66: Nit: these two lines could probably be simplified to `{{ label || 'Select an option' }}` without needing any `<span>` tags.

es-ds-components/components/es-dropdown-select.vue
```diff
@@ -0,0 +1,242 @@
...
+</script>
+
+<template>
+    <div
+        class="input-wrapper justify-content-end"
+        :class="$attrs.class">
+        <label
+            :id="labelId"
+            :for="id"
+            :class="{ 'sr-only': label === '' }"
```
ericdouglaspratt, line 63: Nit: probably safe to do `!label` to make it even simpler.

ericdouglaspratt (dismissed):
Nice work! Two optional things but otherwise good to go.

ericdouglaspratt (approved):
Let's go 🚀

PR #1647, opened 2025-08-11
feat: IDR-1606 create es-toggle

oli3iaseds:
### 🔗 Linked issue

https://energysage.atlassian.net/browse/IDR-1606

### ❓ Type of change

### 📚 Description

- Create toggle component, following [Figma](https://www.figma.com/design/G3Zq89yg3vek9b4JrnlSYn/ESDS-2.0?node-id=16905-1450&t=IntGQ6XQKOkFkMlE-0)
- Added a docs page

### 🥼 Testing

- Tested locally for correct functionality in all states
- Note: WAVE browser gave `Orphaned form label` warning, but Safari VoiceOver confirmed that switch & label were correctly associated

- General feedback

### 📝 Checklist

  - [x] navigating by keyboard

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -100,6 +100,14 @@ jobs:
               working-directory: es-ds-components
               run: npm install
 
+            - name: Commit package-lock.json changes
+              run: |-
+                  git config user.name 'github-actions[bot]'
+                  git config user.email 'github-actions[bot]@users.noreply.github.com'
+                  git add .
+                  git commit -m "chore: update es-ds-components package-lock.json"
+                  git push
+
             - name: Publish Packages
               working-directory: es-ds-components
               id: publish-packages
@@ -107,6 +115,44 @@ jobs:
                   NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
               run: |
                   npm publish
+
+    release-es-ds-docs:
+        permissions:
+            contents: read
+        runs-on: ubuntu-latest
+        needs:
+            - release-please
+            - release-es-ds-styles
+            - release-es-ds-components
+        if: always() && !cancelled() && !failure() && needs.release-please.outputs.es-ds-docs--release_created
+        steps:
+            - name: Check out repository code
+              uses: actions/checkout@v4
+
+            - name: Get nodenv version 🔌
+              id: nodenv
+              shell: bash
+              run: echo "node-version=$(cat .node-version)" >> $GITHUB_OUTPUT
+
+            - name: Set up Node@${{ steps.nodenv.outputs.node-version }} 🏗
+              uses: actions/setup-node@v4
+              with:
+                  node-version: '${{ steps.nodenv.outputs.node-version }}'
+                  registry-url: 'https://registry.npmjs.org'
+                  cache: npm
+                  cache-dependency-path: '**/package-lock.json'
+
+            - name: Install dependencies
+              working-directory: es-ds-docs
+              run: npm install
+
+            - name: Commit package-lock.json changes
+              run: |-
+                  git config user.name 'github-actions[bot]'
+                  git config user.email 'github-actions[bot]@users.noreply.github.com'
+                  git add .
+                  git commit -m "chore: update es-ds-docs package-lock.json"
+                  git push
 
     deploy:
         uses: ./.github/workflows/deploy.yml
--- a/es-ds-components/package-lock.json
+++ b/es-ds-components/package-lock.json
@@ -9,7 +9,7 @@
             "version": "3.5.0",
             "license": "MIT",
             "dependencies": {
-                "@energysage/es-ds-styles": "^3.2.1",
+                "@energysage/es-ds-styles": "^3.3.1",
                 "@nuxt/eslint": "^1.0.1",
                 "@nuxtjs/google-fonts": "^3.2.0",
                 "@vuelidate/core": "^2.0.3",
@@ -584,8 +584,8 @@
         },
         "node_modules/@energysage/es-ds-styles": {
             "version": "3.3.1",
-            "resolved": "https://registry.npmjs.org/@energysage/es-ds-styles/-/es-ds-styles-3.2.1.tgz",
-            "integrity": "sha512-vepDJDpRPNewjsYNziViGHycpHKLoI4GF6VYaOnzOttIRt90QKjBX4dHg99ZB2qiWr0AX2qtslG2TzWjcBXrvA==",
+            "resolved": "https://registry.npmjs.org/@energysage/es-ds-styles/-/es-ds-styles-3.3.1.tgz",
+            "integrity": "sha512-jN+3/WTQ9yIwxZTMRyNZ2eVj9NuVWcNOhBRJIOA31/5DOF5PWcK1RJYrKP4W2aKh6kUQcYMfdmWJTZ4LQo8IDg==",
             "license": "MIT"
         },
         "node_modules/@es-joy/jsdoccomment": {
--- a/es-ds-docs/package-lock.json
+++ b/es-ds-docs/package-lock.json
@@ -9,8 +9,8 @@
             "version": "3.5.0",
             "hasInstallScript": true,
             "dependencies": {
-                "@energysage/es-ds-components": "^3.2.3",
-                "@energysage/es-ds-styles": "^3.2.1",
+                "@energysage/es-ds-components": "^3.5.0",
+                "@energysage/es-ds-styles": "^3.3.1",
                 "@nuxt/image": "^1.8.0",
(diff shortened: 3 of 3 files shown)


nathanielwarner (changes requested):
This component looks quite nice!

However there's one accessibility problem that needs to be fixed. There is no label attached to these switches. There should be a mechanism to attach a label to the PrimeVue component, following [PrimeVue's guidance](https://v3.primevue.org/inputswitch/#accessibility), and our documentation examples should reflect this usage. The particulars of this depend on your needs- perhaps there'll always be an adjacent label, which you can pass in the `id` of, utilizing the `aria-labelledby` approach shown by PrimeVue.

Feel free to reach out if you have any questions or are not sure how to proceed.

Side note: The WAVE browser extension (one of the checklist steps in the PR template) reported this error.

es-ds-docs/pages/molecules/toggle.vue
```diff
@@ -0,0 +1,67 @@
+<script setup lang="ts">
+const propTableRows = ref([['disabled', 'boolean', 'false', 'Disables the toggle.']]);
```
hroth1994, line 2: nit: capitalize `Boolean` here to match the props tables from other components

es-ds-docs/pages/molecules/toggle.vue
```diff
@@ -0,0 +1,67 @@
...
+            </div>
+            <div>
+                <es-toggle
+                    v-model="disabledChecked"
+                    disabled />
+            </div>
+        </div>
+
+        <div class="my-500">
+            <h2>EsToggle Props</h2>
```
hroth1994, line 57: nit: un-capitalize `props` to match "Sentence case" used throughout the design system

hroth1994 (changes requested):
Thanks for working on this component! Just 2 small changes I noticed:

es-ds-components/components/es-toggle.vue
```diff
@@ -0,0 +1,170 @@
...
+                <p
+                    v-if="hasValue"
+                    id="es-toggle-value"
+                    class="es-toggle-value">
+                    {{ model ? onValue : offValue }}
+                </p>
+            </div>
+            <div v-if="$slots.text">
+                <slot name="text" />
+            </div>
```
gigiwargin, line 61: If you check out the current switch usage in es-onboarding, you'll note that the "description" is linked to the toggle in such a way that when I click on the text, it switches the toggle on and off. I think that for accessibility and usability purposes it would make sense to have that same behavior here.

https://github.com/user-attachments/assets/0b99ac57-0b6a-4e08-9c65-511c276c7cfc

gigiwargin (commented):
Few comments/questions inline, now that this component is more complex! Thanks for rolling with the punches

es-ds-components/components/es-toggle.vue
```diff
@@ -0,0 +1,158 @@
...
+                    v-model="model"
+                    :disabled="disabled"
+                    :aria-label="ariaLabel"
+                    :aria-labelledby="label ? 'es-toggle-value' : undefined"
+                    class="es-toggle">
+                    <switch-thumb class="es-toggle-thumb" />
+                </switch-root>
+                <p
+                    v-if="label"
+                    id="es-toggle-value"
```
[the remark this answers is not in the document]
nathanielwarner: Also, is this visible label something that was prompted just by the accessibility requirement? Would we use this in practice? Don't the labels usually appear to the side?

es-ds-components/components/es-toggle.vue
```diff
@@ -0,0 +1,158 @@
...
+            class="d-flex"
+            :class="{ 'has-text': $slots.text }">
+            <div
+                class="d-flex flex-column align-items-center"
+                :class="{ 'has-value': label }">
+                <switch-root
+                    v-model="model"
+                    :disabled="disabled"
+                    :aria-label="ariaLabel"
+                    :aria-labelledby="label ? 'es-toggle-value' : undefined"
```
nathanielwarner, line 41: Do we not want to allow the user to pass in the id of a label element that's adjacent to the switch?

es-ds-docs/pages/molecules/toggle.vue
```diff
@@ -0,0 +1,110 @@
...
+            <es-toggle class="mb-100" />
+            <es-toggle
+                v-model="checked"
+                class="mb-100" />
+        </div>
+        <div class="my-500">
+            <h2>With label</h2>
+            <es-toggle
+                class="mb-100"
+                label="ON" />
```
nathanielwarner, line 52: The purpose of the toggle's label is to describe _what_ we're toggling, not the state of the toggle itself, which is already handled by the screen reader out of the box. See [PrimeVue's example](https://v3.primevue.org/inputswitch/#accessibility), where it's used to toggle "Remember me" functionality.

es-ds-components/components/es-toggle.vue
```diff
@@ -0,0 +1,158 @@
...
+                    class="es-toggle-value">
+                    {{ label }}
+                </p>
+            </div>
+            <div
+                v-if="$slots.text"
+                :class="{ 'cursor-pointer': !disabled }"
+                @click="!disabled && (model = !model)">
+                <slot name="text" />
+            </div>
```
gigiwargin, line 57: I think we may actually want this to look something like this, given `es-toggle-id` is the `id` attribute on `<switch-root>`. I pulled this example from the code in the [first example from the Reka UI docs](https://reka-ui.com/docs/components/switch#switch). This will provide the same functionality as the `@click` event, and more (e.g. hovering, focusing, etc).
```suggestion
            <label
                v-if="$slots.text"
                for="es-toggle-id">
                <slot name="text" />
            </label>
```
This will require each instance of `es-toggle` to have a unique `id` attribute that gets defined or passed in, which I think Nathaniel called out in an earlier review!

es-ds-components/components/es-toggle.vue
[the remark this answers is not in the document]
gigiwargin: +1

I think it makes sense to keep the slot for the label/description shown in the screenshot, but the label underneath the switch doesn't seem necessary yet. No point in adding something that may never be used.

gigiwargin (commented):
Few comments inline. You've done a great job with this!! 🥳 ✨

es-ds-docs/pages/molecules/toggle.vue
oli3iaseds: Removing this label after talking w Gigi since it's not required for our current use cases

nathanielwarner (approved):
LGTM! 🚀 Thank you for your hard work on this. It's awesome that our design system has a toggle now!

es-ds-components/components/es-toggle.vue
```diff
@@ -0,0 +1,133 @@
...
+        opacity: 0.5;
+
+        :deep(*) {
+            opacity: inherit;
+        }
+    }
+}
+
+.es-toggle-label {
+    margin-top: 8px;
```
ericdouglaspratt, line 62: Worth throwing a `user-select: none` in here so this text highlighting doesn't happen if you click too quickly?

https://github.com/user-attachments/assets/b918944a-2491-4065-a28d-db52e351b944

es-ds-components/components/es-toggle.vue
```diff
@@ -0,0 +1,133 @@
...
+defineProps({
+    disabled: {
+        type: Boolean,
+        default: false,
+        description: 'Disables the toggle.',
+    },
+});
+
+// Workaround: Reka UI components don't play nice with the Nuxt 3.16+ dev server side rendering while using NPM link
+const wrappingComponent = import.meta.dev ? resolveComponent('ClientOnly') : 'span';
```
ericdouglaspratt, line 18: The fallback element here might need to be a `'div'` rather than a `'span'` because I think it's technically invalid HTML to have a `<div>` (block element) inside of a `<span>` (inline element).

es-ds-components/components/es-toggle.vue
```diff
@@ -0,0 +1,133 @@
...
+const model = defineModel<boolean>({
+    default: false,
+});
+
+defineProps({
+    disabled: {
+        type: Boolean,
+        default: false,
+        description: 'Disables the toggle.',
+    },
```
ericdouglaspratt, line 14: Totally optional since this props definition works and we do use this more options-API-like syntax in a lot of ESDS components still, but an alternative that's more TypeScript-y (and potentially more concise when there are a lot of props) is below. The presence or absence of the `?` indicates whether or not it's required, and you can use an actual TypeScript interface to define the types of the props.

```
interface Props {
    disabled?: boolean;
}
withDefaults(defineProps<Props>(), {
    disabled: false,
});
```

ericdouglaspratt (changes requested):
Looks lovely, nice work! Just a couple small things.

PR #1653, opened 2025-08-21
fix: define value for dropdown option

hroth1994:
### ❓ Type of change

### 📚 Description

In order to get just the value from the selected option in the dropdown, add `optionValue` prop to `dropdown` that is defined as `value` when the options prop is an array of objects. Also add `v-bind="$attrs"` in case there is a need to pass other props in the future.

### 🥼 Testing

Tested on http://localhost:8500/molecules/dropdown-select
- Non-object examples still work as expected
- Object example returns the value without needing to extract with `.value`

- General

### 📝 Checklist

  - [ ] navigating by keyboard

### DIFF
--- a/es-ds-components/components/es-dropdown-select.vue
+++ b/es-ds-components/components/es-dropdown-select.vue
@@ -85,6 +85,7 @@ const hide = () => {
             :focus-on-hover="false"
             :model-value="modelValue"
             :option-label="options.length > 0 && typeof options[0] === 'object' ? 'label' : undefined"
+            :option-value="options.length > 0 && typeof options[0] === 'object' ? 'value' : undefined"
             :options="options"
             :placeholder="placeholder"
             :pt="{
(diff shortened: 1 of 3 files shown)


ericdouglaspratt (changes requested):
Looks good but in testing this I encountered one issue we missed during initial implementation of EsDropdownSelect - I brought up the local dev server in my browser, clicked "Dropdown select" in the nav, and the entire page reloaded and stayed on the index page with this message in the server console:

```
ℹ ✨ new dependencies optimized: primevue/dropdown                                                             10:22:44 AM
ℹ ✨ optimized dependencies changed. reloading  
```

Can we add `'primevue/dropdown'` to the list of components in the `vite.optimizeDeps.include` array in the es-ds-components nuxt.config.ts to avoid this issue?

ericdouglaspratt (approved):
Nice!

PR #1676, opened 2025-09-17
fix: pass toggle attrs to inner div

hroth1994:
### ❓ Type of change

### 📚 Description

Our components that use Reka UI come with an interesting caveat - client vs. server-side rendering is different based on the following code:
```
// Workaround: Reka UI components don't play nice with the Nuxt 3.16+ dev server side rendering while using NPM link
const wrappingComponent = import.meta.dev ? resolveComponent('ClientOnly') : 'div';
```
In order to get classes to pass to the inner div that is only generated on the server side, `v-bind="$attrs"` must be added and `inheritAttrs: false` must be set so that classes aren't applied to the outer div

### 🥼 Testing

Tested by adding a toggle to the docs page that passes utility and custom classes and running `npm run generate`. I removed it after testing because it is a rather specific example of styling for HomeScore

Screenshot 2025-09-17 at 9 43 34 AM

- General

### 📝 Checklist

  - [x] navigating by keyboard

### DIFF
--- a/es-ds-components/components/es-toggle.vue
+++ b/es-ds-components/components/es-toggle.vue
@@ -1,5 +1,9 @@
 <script setup lang="ts">
 import { SwitchRoot, SwitchThumb } from 'reka-ui';
+// Prevents attributes from being applied to first <div>
+defineOptions({
+    inheritAttrs: false,
+});
 
 const switchId = useId();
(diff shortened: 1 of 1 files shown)


nathanielwarner:
👀

PR #1734, opened 2026-04-24
fix: update dependencies, update node, follow best practices

nathanielwarner:
### ❓ Type of change

### 📚 Description

- Updates node version to latest LTS
- Updates dependencies by choosing new major versions for ESLint and TypeScript
- Remove overrides
- Update tsconfigs to follow [recommended approach from Nuxt](https://nuxt.com/docs/4.x/directory-structure/tsconfig)
- Remove some files which aren't used anymore, such as those from Backstage
- In the Nuxt Layer (`es-ds-components`), put all dependencies in `devDependencies` since they're not actually supposed to be installed downstream, following the [example from Nuxt](https://github.com/nuxt/starter/blob/layer/package.json)

### 🥼 Testing

- Tested downstream with local linking to these branches which have analogous changes:
   - https://github.com/EnergySage/es-storyblok-shared/compare/main...CEO-561-update-deps
   - https://github.com/EnergySage/es-cms-pages/compare/develop...CEO-561-update-deps
   - Checked dev server and static generation
   - I still occasionally see errors in the console related to TypeScript files, when loading the dev server for the first time. These don't seem to have an effect, and I haven't been able to resolve them. I did adjust the `make link` scripts downstream to run `npm rebuild` after doing the local linking, which may have helped a bit.
- Note that the first time you load up this branch you'll probably need to run `nodenv install`

- General

### 📝 Checklist

  - [x] navigating by keyboard


es-ds-components/nuxt.config.ts
```diff
@@ -92,13 +92,15 @@ export default defineNuxtConfig({
                 'primevue/tabview',
                 'primevue/tooltip',
                 'reka-ui',
+                '@phosphor-icons/vue',
+                'maska/vue',
```
ericdouglaspratt: I get this warning on dev server startup: `WARN  Unresolvable optimizeDeps.include entries: maska/vue`

ericdouglaspratt (changes requested):
Looks good, dev server and generate work locally!

Just one thing I saw, plus after my install there was a one line update to the package-lock.json in es-ds-components.

es-ds-components/nuxt.config.ts
nathanielwarner: Addressed in [e8f96ed](https://github.com/EnergySage/es-ds/pull/1734/commits/e8f96ed43fa35e887b768a938758f5671c7e9351)

nathanielwarner:
@ericdouglaspratt could you let me know what you're seeing in the package-lock? And also confirm that you followed the instruction in the PR description to run `nodenv install`? (and that running `node --version` returns the version in the `.node-version` file)

ericdouglaspratt:
@nathanielwarner it just removed the `"dev": true,` from `"node_modules/@parcel/watcher-wasm/node_modules/napi-wasm": {` - and yep the Node version looks right and is all installed

nathanielwarner:
@ericdouglaspratt that's interesting- I've seen this sort of thing sporadically and don't know what would be causing it. I wiped out my `node_modules` folders and tried `make install` again, and it's not changing the package-lock.

ericdouglaspratt (approved):
Ok! Looks good

PR #1746, opened 2026-05-11
feat: add EsMenuBar, EsSkipToContentLink, EsColorBand

ericdouglaspratt:
### ❓ Type of change

### 📚 Description

- Added EsMenuBar and child components
- Fixed issues with EsStickyBar to enable supporting the nav
- Added EsColorBand
- Added EsSkipToContentLink and implemented it on the ESDS docs site
- Added `useEsdsEvents` global events composable, using the `mitt` library, to allow the components to talk to each other
- Updated site navigation example page showing the complete suite of components and how they interact (minus the featured article piece which IMO doesn't yet belong in ESDS)
- Added fan and person-circle icons

### 🥼 Testing

- Tested locally on docs page and site navigation example
- Tested with VoiceOver and keyboard navigation
- Tested in Chrome, Firefox, Edge, Safari

- Overall

### 📝 Checklist

  - [x] navigating by keyboard


hroth1994:
👀

es-ds-components/app/components/es-menu-bar.vue
```diff
@@ -0,0 +1,514 @@
...
+                    }
+                }
+
+                .es-menu-bar-flyout-pane {
+                    margin: 0 auto;
+                    max-width: map.get(variables.$container-max-widths, xxl);
+
+                    /* at xxl breakpoint only, match es-container padding */
+                    @include breakpoints.media-breakpoint-up(xxl) {
+                        padding-left: variables.$grid-gutter-width * 0.5 !important;
```
hroth1994, line 414: nit: could be combined to `padding-x: variables.$grid-gutter-width * 0.5 !important;`

es-ds-components/app/components/es-menu-bar.vue
```diff
@@ -0,0 +1,514 @@
...
+        opacity: 0;
+        transform: translateX(calc(-1 * var(--es-menu-bar-flyout-translate)));
+    }
+    to {
+        opacity: 1;
+        transform: translateX(0);
+    }
+}
+
+@keyframes enterFromRight {
```
hroth1994, line 151: Since 5 of these keyframes are the exact opposites of another, it would reduce the amount of code if they were removed and `reverse` was used instead. 
For example, instead of:
`animation: menuClose var(--es-menu-bar-open-close-duration) $open-close-timing-function;`
you could reuse `menuOpen` with:
`animation: menuOpen var(--es-menu-bar-open-close-duration) $open-close-timing-function reverse;`

hroth1994 (commented):
This looks really nice! The animations are smooth and the components are laid out in a reusable way. I can tell you put alot of effort into this and it certainly paid off 😃 
I just have a couple suggestions for simplifying the CSS, take them or leave them!

es-ds-components/app/components/es-menu-bar.vue
ericdouglaspratt: 🤔 I don't think `padding-x` is a valid CSS property name. I just tried it and it said no

Screenshot 2026-05-18 at 2 41 33 PM

es-ds-components/app/components/es-menu-bar.vue
ericdouglaspratt: Hmm that causes everything to get messed up somehow. Might be something to do with the `cubic-bezier` timing function not being symmetric- since the enter/exit animations at least need to run in sync. But the behavior below looks weird enough that there might be even something else going on, so I'm inclined to leave as is and not mess with it.

https://github.com/user-attachments/assets/9ec7ebd1-d799-419b-a68b-d9ea45f48fcc

ericdouglaspratt: Ohh I just tried it with `menuOpen` as well; I think what's happening is that Reka UI's automatic detection of whether there's an entrance or exit animation on the element isn't finding the animation when it has `reverse` on it, so it's just unmounting it immediately and not allowing the animation to happen.

https://github.com/user-attachments/assets/29883d45-2262-44a4-9c7e-d74c6fbe6326

es-ds-components/app/components/es-menu-bar.vue
hroth1994: oops I was thinking in utility classes, I meant something like `padding: 0 variables.$grid-gutter-width * 0.5 !important;` but it's fine to leave it the way you have it

es-ds-components/app/components/es-menu-bar.vue
hroth1994: Interesting 🤔 yep let's leave it as is

hroth1994 (approved):
Let's goooo! 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #1753, opened 2026-05-21
fix: make local dev server fast again

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CEO-624

### ❓ Type of change

### 📚 Description

- Spent four hours with Claude diagnosing the root cause of the issue and eliminating non-causes, including by running CPU profiling on the dev server via `NODE_OPTIONS="--cpu-prof --cpu-prof-dir=./.profiles --cpu-prof-interval=1000" npx nuxi dev` (which outputs useful files in a `.profiles` directory and is what ultimately led us to the correct solution)
- Root cause was [Nuxt 4.4.0 adding "Better Import Protection"](https://github.com/nuxt/nuxt/releases/tag/v4.4.0) in https://github.com/nuxt/nuxt/pull/34454 via the `impound` library
- Patching that library to short-circuit its checks when in development mode fixed the dev server slowness issue, while ensuring that its import checks (which prevent server code, potentially containing secrets, from being included in the client-side bundle, which is a non-issue in this repo anyway since there is no server code) will still be run during integration and production builds
- Also added a markdown file right next to the patch file that explains everything

### 🥼 Testing

- Tested locally by starting up dev server and seeing that second and third page loads are fast rather than 40s

- Overall

### 📝 Checklist

  - [ ] navigating by keyboard

### DIFF
--- a/es-ds-docs/patches/impound+1.1.5.patch
+++ b/es-ds-docs/patches/impound+1.1.5.patch
@@ -1,0 +1,19 @@
+diff --git a/node_modules/impound/dist/index.mjs b/node_modules/impound/dist/index.mjs
+index 1d3223f..3f4f5ee 100644
+--- a/node_modules/impound/dist/index.mjs
++++ b/node_modules/impound/dist/index.mjs
+@@ -282,6 +282,14 @@ const ImpoundPlugin = createUnplugin((globalOptions) => {
+ 	});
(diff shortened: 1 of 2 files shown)


hroth1994 (approved):
Confirmed local dev server runs way faster!

PR #1778, opened 2026-07-27
fix: EsCarousel autoplay no longer interrupts screen readers

ericdouglaspratt:
### 🔗 Linked issue

- https://energysage.atlassian.net/browse/CEO-676

### ❓ Type of change

### 📚 Description

- EsCarousel has been moved off of PrimeVue and now uses Embla Carousel
- Major EsCarousel accessibility fixes and improvements:
     - Autoplay no longer reads out each slide live in screen reader mode, interrupting navigation through the rest of the page
     - The reduce motion setting is now respected
     - It now reads to screen readers as an actual carousel with better semantics
- Circular carousels now loop seamlessly rather than occasionally reversing the entire set of slides back to the start position
- Swiping now incorporates momentum, and drag now exactly stays with the user's tap/pointer location rather than just being a directional indicator to start a full swipe animation
- Added "No arrows" example to EsCarousel docs page
- EsCarousel examples now have empty alt text on their images so extraneous image data is not read out by screen readers for what is decorative only

### 🥼 Testing

- Tested locally at http://localhost:8500/organisms/carousel
- Also linked to a downstream repo and tested the carousel on those pages and it functioned quite well with no aria-live readout and good core functionality as well

- Overall

### 📝 Checklist

  - [x] navigating by keyboard


hroth1994:
👀

es-ds-components/app/components/es-carousel.vue
```diff
@@ -1,29 +1,37 @@
...
-     - prop to position the arrows at the bottom two corners of a full-width slide, like homepage
+    EsCarousel is built on Embla Carousel (https://www.embla-carousel.com/), replacing the
+    previous PrimeVue implementation. Embla is SSR-friendly and gives us full control over
+    the markup, which lets us make the component fully accessible.
+
+    TODO (design):
+     - Add a visible pause/play button when `autoPlay` is on. Today autoplay can only be
+       stopped via the Esc key, which is not reachable in mobile screen readers. A visible
+       control satisfies WCAG 2.2.2, but where it lives and how it looks needs a design pass.
+       Iterating on that with design is out of scope for this ticket (CEO-676).
```
hroth1994: I think most of this comment is useful except for the last sentence, where it seems like Claude is talking to itself.

es-ds-components/app/components/es-carousel.vue
```diff
@@ -1,29 +1,37 @@
...
+// constants that contribute to dots and control sizing/spacing
 const DOT_SIZE = 14;
 const DOT_SPACING = 16;
 const ARROW_BUTTON_PADDING = 8;
 
-// the number of "dots" apart the arrows should be when dots are hidden
-const ARROW_SPACING_WHEN_NO_DOTS = 1;
-
 interface IProps {
+    /**
```
hroth1994: This comment seems unnecessary since the props are detailed on the docs page.

es-ds-components/app/components/es-carousel.vue
```diff
@@ -73,490 +81,448 @@ const BREAKPOINTS = {
...
-        },
-    ];
-});
+const scrollPrev = () => emblaApi.value?.scrollPrev();
+const scrollNext = () => emblaApi.value?.scrollNext();
+const scrollTo = (index: number) => emblaApi.value?.scrollTo(index);
 
-const autoplayInterval = ref(props.autoPlay ? props.autoPlayInterval : 0);
+const stopAutoplay = () => {
+    // stop carousel when user presses Escape key, in lieu of a pause button
```
hroth1994: This comment should probably be moved to `onEscapeKeyup`

es-ds-docs/app/pages/organisms/carousel.vue
hroth1994: Not entirely relevant to this PR, but this CSS isn't being used and can be removed

es-ds-components/app/components/es-carousel.vue
```diff
@@ -73,490 +81,448 @@ const BREAKPOINTS = {
...
+    auto-advancing content is motion in its own right, so the preference should stop it entirely
+    rather than merely removing the slide animation. (Manual paging stays instant via `duration`.)
+
+    It is a plugin, added at creation time. stopOnInteraction is false so it matches the previous
+    behavior of running until the user presses Esc.
+*/
+const autoplayEnabled = props.autoPlay && !prefersReducedMotion.value;
+const autoplayPlugins = autoplayEnabled ? [Autoplay({ delay: props.autoPlayInterval, stopOnInteraction: false })] : [];
+
+const [emblaRef, emblaApi] = emblaCarouselVue(emblaOptions.value, autoplayPlugins);
```
hroth1994: Based on the [Embla source code](https://github.com/davidjerleke/embla-carousel/blob/a4f9744c0503cef470001215569e94ed6544543f/packages/embla-carousel-vue/src/components/useEmblaCarousel.ts#L56), if you pass `emblaOptions` as a ref here then it will set up its own watcher that calls reInit() automatically when options change. So the `watch` below on `emblaOptions` would be unnecessary if it's passed as a ref (without `.value`) here.

hroth1994 (changes requested):
Nice job, this is definitely an accessibility improvement! I added a few suggested changes for code quality.

es-ds-components/app/components/es-carousel.vue
ericdouglaspratt: Good catch! It looks like that also reduces computation because it checks - in a deeply nested way - if the options are equal before doing anything, which I don't think was happening before.

hroth1994 (approved):
All my comments have been addressed, nice job! 🎠

